PR #4, opened 2017-04-17
Some minor suggested changes

bep:
Great theme!


layouts/index.html
```diff
@@ -2,8 +2,10 @@
   <main class="cf ph3 ph5-l pv3 pv4-l f4 tc-l center measure-wide lh-copy mid-gray">
     {{ .Content }}
   </main>
-  {{/* Create a variable to use the same section twice. Note that we've included the date here. */}}
-  {{ $section := where .Data.Pages.ByDate.Reverse "Section" "post" }}
+  {{/* For Hugo 0.20 this will default to the section with the most number of pages. */}}
```
digitalcraftsman: Since this functionality was introduced with Hugo v0.20 don't forget to bump the minimal Hugo version in the `theme.toml` accordingly.

bep: While I haven't tested it on Hugo 0.18, it should work just fine. That is kind of some of the point of it.

bep: That said, if this is like a "demo theme for Hugo", there are other places I see some constructs that could be upgraded (the alternate RSS feed link in the header being one).

digitalcraftsman: > While I haven't tested it on Hugo 0.18, it should work just fine. That is kind of some of the point of it.

If a certain attribute doesn't exist that someone tries to access (i.e. `.Site.Params.mainSections` in versions prior to v0.20) does this usually raises an error. How could `default` handle this?

bep: But testing this, I notice that the templates overall requires a "post" section, so the above isn't enough to make this work in general.

budparr: @bep 
- what's the issue with the RSSLink? I took that from a PR of yours from the [AMP theme](https://github.com/wildhaber/gohugo-amp/pull/33/files). I really want to push this theme to be exemplary, so do let me know about those other places if you have time.

- I'm using `post` specifically to create examples of using layouts. Open to other ideas.

- From the conversation, I'm not sure if I'm to merge the `mainSections` bit? My biggest goal with that section on the home page is to show how to use a variable and use `first` and `after` - if there are too many things might make it harder to understand. What do you all think?

bep: Nah, this PR is mostly pointless when the section is hard wired to "post". My point about `RSSLink` is that is `RSSLink` and not `RSSlink`-- it used to be both, bot no only the former.

bep: My other point about `RSSLink` may be not using it at all, but use the `.AlternativeFormats` maybe ...

budparr: Well thinking about it, I think what you have is the better, more flexible way to present the home page. If I just use a partial instead of a content view there it will work (and I just spotted another error I made on the second section). So I'm going to take this and work from there.

PR #13, opened 2017-04-19
Fix featured image logic

bep:


layouts/partials/summary-with-image.html
```diff
@@ -3,7 +3,8 @@
     <div class="flex flex-column flex-row-ns">
       {{ if .Params.featured_image }}
         <div class="pr3-ns mb4 mb0-ns w-100 w-40-ns">
-          <img src="{{ .Params.featured_image }}" class="db" alt="image from {{ .Title }}">
+          {{ $imgURL := (trim .Params.featured_image "/") | absURL }}
```
bep: @budparr but you may consider the Param method here, too. I just made it work.

budparr:
Can you tell me why you removed the `featured_image` key from existing posts where there is no image? Reason I ask is that it's in the archetype so will be created automatically in the future.

bep:
Because the `Param` func does a lookup in a map and finds a value (empty string in this case), and does not look further (i.e. the site params). But I don't know what's the "correct" behaviour here.

budparr:
@bep I'm going to close this and re-write them myself, because I realize now I don't want the default behavior on posts (because the default should be no hero), and It's a good exercise for me anyway. Really appreciate you taking the time to do this. Thanks!

PR #40, opened 2017-07-23
Make Hero image work out-of-the box

bep:
* using Hero sample image in the theme's /static folder
* fall back to site config
* fall back to hard coded default if none set in page or site params

See https://github.com/gohugoio/hugoDocs/pull/76

I have started a rework of the Hugo Quick Start, and it would be good to get a nice looking site without too much config twiddling.


budparr:
Thanks, @bep I had this once and reverted because it implied that a user wanted a hero image on every page, creating a lot of redundant default hero images on the site. I can probably deal with that in another way and add this for the sake of the quick start.

PR #49, opened 2017-09-20
add support for google analytics

bdashrad:
add native hugo support for google analytics (https://gohugo.io/templates/internal/#google-analytics)

### DIFF
--- a/layouts/partials/site-footer.html
+++ b/layouts/partials/site-footer.html
@@ -4,5 +4,6 @@
     &copy; {{ now.Format "2006" }} {{ .Site.Title }}
   </a>
   {{ partial "social-follow.html" . }}
+  {{ template "_internal/google_analytics.html" . }}
   </div>
 </footer>


bdashrad:
I chose to use the standard template over the "async" one, based on which code google currently provides for analytics setup.

budparr:
Hey @bdashrad I'm pretty sure it's best practice to load GA async in the head of the site, which we have here: https://github.com/budparr/gohugo-theme-ananke/blob/master/layouts/_default/baseof.html#L34

In a (near) future version of Ananke I'm going to make it easier to customize what goes in the head.

Thanks!

bdashrad:
oh for sure! i swear i searched and didn't see it! thanks for the great work.

bdashrad:
I figured out the problem. I didn't realized there was an extra variable required (`HUGO_ENV` or .Site.Params.env`) to enable it. Hugo seems to handle this internally (doesn't add the google analytics block if using `hugo serve`), so I didn't realize I needed to add this extra variable.

budparr:
Ah, I'm glad you pointed that out. I'll be sure to add something to the readme on that. Thanks!

PR #52, opened 2017-09-26
Adding LinkedIn Icon #27

adambkaplan:

PR #51, opened 2017-09-22
Social icon updates

bdashrad:
- Add LinkedIn
- Fix github svg spacing
- Update colors to ones from brandcolors.net

PR #53, opened 2017-09-26
Fix the social icon spacing

bdashrad:
This will actually fix the spacing, per the comments from @adambkaplan and @budparr in #51 and #52.


budparr:
I think we crossed paths on this. I fixed it just after pulling in the original PR. Reopen if not.

bdashrad:
Ah, perfect. didn't even see your addition. Thanks for a great theme and quick response.

Issue #68, opened 2017-10-31
Parmeterize number of recent posts.


DeveloperDavo:
Hello, 

I think it would be nice to parameterize the number of recent posts to display in `index.html` instead of hardcoding to `3` in `{{ range (first 3 $section) }}`. Something like:
```
[params]
  recent_posts_number = 5  
```
in `config.toml` to replace `3` in `index.html`.

Let me know if you would like me to submit a pull request for this.

Thanks,

David

budparr:
Sure, that's a great idea. PR is welcome!

budparr:
solved in #69

PR #69, opened 2017-11-01
#68|Parmeterize number of recent posts in index.html

DeveloperDavo:


DeveloperDavo:
Let me know if you would also like me to parameterize the number of more posts as well. Something like:
```
[params]
  more_posts_number = 4  
```

layouts/index.html
```diff
@@ -20,20 +20,22 @@ <h1 class="flex-none">
           </h1>
         {{ end }}
 
+      {{ $n_posts := .Site.Params.recent_posts_number | default 3 }}
```
budparr: This should be `$.Param "recent_posts_number"` [Param](https://gohugo.io/functions/param) allows users to set the variable at the page level or config.

budparr:
Thank you, @DeveloperDavo I just had the one comment.

budparr:
Thanks again!

PR #76, opened 2018-01-08
Fix pagination

rcoh:
Pagination doesn't actually work -- it just shows the pagination nav and the content repeats over and over. 

This commit fixes pagination.


budparr:
Funny, @rcoh I don't recall what my intention was there. I have pagination on the post template. At any rate, thanks for the PR.

PR #83, opened 2018-02-05
Add support for logo in main Site-header

Titan-C:
This theme is great. I though of having a logo in the main site. Are you interested in supporting this feature?

### DIFF
--- a/layouts/partials/site-header.html
+++ b/layouts/partials/site-header.html
@@ -6,9 +6,13 @@
     <div class="bg-black-60">
       {{ partial "site-navigation.html" .}}
       <div class="tc-l pv4 pv6-l ph3 ph4-ns">
+      {{ if .Site.Params.logo }}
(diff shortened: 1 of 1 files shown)


budparr:
Hey, @Titan-C Can you show me an example of this? I think that putting an image over an already existing image could be problematic for most users. 

I just threw a small logo in to test and this is what I came up with:

screen shot 2018-02-05 at 11 17 54 am


Might be a situation where it's best that you just create a local copy for your needs rather than make it a hard-coded thing for all users.

Titan-C:
I use a very simple svg image for the logo. I like that the logo is on top, but not part of the image. It behaves like the title and is responsive.

image

Sure I'll keep this in my local copy if there is no interest.

budparr:
That looks great! But yeah, I think for most users it would be problematic. Thanks!

PR #87, opened 2018-02-26
Fixing reference to site description from [params] section of config.toml

toddmeinershagen:
Fixing reference to site description from the [params] section of the config.toml in the site-header.html document.

Before this, the theme would not display the description.

### DIFF
--- a/layouts/partials/site-header.html
+++ b/layouts/partials/site-header.html
@@ -1,4 +1,5 @@
 {{ $featured_image := .Param.featured_image | default "images/gohugo-default-sample-hero-image.jpg" }}
+{{ $description := .Param.description | default "" }}
 {{ if $featured_image }}
   {{/* Trimming the slash and adding absURL make sure the image works no matter where our site lives */}}
   {{ $featured_image := (trim $featured_image "/") | absURL }}
@@ -9,7 +10,7 @@
         <h1 class="f2 f-subheadline-l fw2 white-90 mb0 lh-title">
           {{ .Title | default .Site.Title }}
         </h1>
-        {{ with .Param.description }}
+        {{ with $description }}
           <h2 class="fw1 f5 f3-l white-80 measure-wide-l center mt3">
             {{ . }}
           </h2>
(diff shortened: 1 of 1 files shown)


budparr:
If we used `$.Param "description"` it would fall back on the site description, if that exists. Do you think that would be preferable to an empty default?

toddmeinershagen:
.Param "description" and .Param.description are the same thing.  So, I
don't think having the first be the default if the second is not set will
produce anything anyway.  That is why I set it to an empty string.

Perhaps we could set it to the default value for the description of the
Ananke theme - a string like "The Ananke theme.".  I'll leave that up to
you to decide.

Hope it helps,


Todd

budparr:
I could be mistaken, but the dollar sign ($) makes it global, picking up a description parameter from config . I'm not sure from an SEO perspective, whether it's better to have duplicate descriptions or none. Nonetheless, I'll have a look at this in a bit.

Really appreciate the contribution!

budparr:
Hey, @toddmeinershagen Can you show me the repo for your site? I'm just went and looked at this and I'm not seeing any problem with the description. It looks like you're referring to the description on the home page or other pages, not the "post" content type, is that right? I just want to troubleshoot some more.

toddmeinershagen:
I am referring to the description that displays on the home page.

My repo is located at:
https://github.com/toddmeinershagen/gohugo-theme-ananke

LimitedEdition1:
Based on the Hugo documentation: https://gohugo.io/functions/param/

A significant difference with .Param "KEY" is inheritance.  This is the basic change I made for the social link pull request. It doesn't break existing user sites but adds much more flexibility to override the config defaults using the page front matter. Either with a specific value or using a value of (empty string or false) to choose nothing instead of the config default.  

This way if you have a lot of pages, and you want a default, set that in the config.toml, and for those pages you want nothing or something different, add to the page front matter. 

I'm only less than a week into using Hugo, so it's very possible I missed something as well. Just thought I'd comment based on what I learned with social links.

regisphilibert:
> .Param "description" and .Param.description are the same thing. 

Nope. The `.Param` page method look for a page param and, if not found, look for the site param of the same key. Only if none of those are found does hugo returns empty/false.

Following #141, we concluded this is a breaking change. 

We're getting back at giving some love to the theme, maybe we could address this concern at a later time.

Issue #97, opened 2018-04-14
Hugo fails to load CSS and Javascript files without baseURL


anddongyangli:
Hugo fails to load the assets in debug mode when the `baseURL` is set to something like `"https://example.com/hugo/"`. A 404 page appears when accessing the css or Javascript file directly via the web browser.

We might want to set absolute URLs for these assets based on the configured `baseURL`.

Thanks.

budparr:
Thanks for pointing this out, @anddongyangli I'll check it out.

anddongyangli:
@budparr  Thanks for addressing this issue.

This issue still exists if the baseURL contains a non-root context path.
For example,  set the `baseURL` in  the configuration file like this:
```

baseURL = "http://www.foobar.com/blog"
```

Starting hugo server in debug mode, the following code
```go
 {{ $script := .Site.Data.webpack_assets.app }}
{{ with $script.js }}
  <script src="{{ absURL (printf "%s%s" "/dist/" .) }}"></script>
{{ end }}
```
will be compiled to
```
<script src="http://localhost:1313/dist/js/app.3fc0f988d21662902933.js"></script>
```
instead of 
```
<script src="http://localhost:1313/blog/dist/js/app.3fc0f988d21662902933.js"></script>
```

I tried with the following code, it works fine.

```go
{{ $script := .Site.Data.webpack_assets.app }}
{{ $baseURL := .Site.BaseURL }}
{{ with $script.js }}
    <script src="{{ printf "%s%s%s"  $baseURL  "/dist/" . }}"></script>
{{ end }}

```

anthonyfok:
Thanks @anddongyangli!  But the fix is simpler: simply removing the leading slash, i.e. changing "/dist/" to "dist/" would do the trick because absURL already takes the baseURL into consideration when it sees a relative path.

On second thought, maybe relURL would work just as well.

PR #103, opened 2018-05-02
Fix asset paths when baseURL has sub-folder

anthonyfok:
Hi @budparr,

https://themes.gohugo.io/theme/gohugo-theme-ananke/, a case where baseURL contains sub-folder, is not rendering correctly.  It appears that absURL behaves differently when the path is prefixed with a `/` slash.  The fix is to simply remove that leading slash.

The documentation at https://gohugo.io/functions/absurl/ does not seem to mention this fact explicitly.  Maybe we should fix the hugoDocs too.

See also #97

Thanks!

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -19,7 +19,7 @@
 
     {{ $stylesheet := .Site.Data.webpack_assets.app }}
     {{ with $stylesheet.css }}
-      <link href="{{ absURL (printf "%s%s" "/dist/" .) }}" rel="stylesheet">
+      <link href="{{ relURL (printf "%s%s" "dist/" .) }}" rel="stylesheet">
     {{ end }}
 
     {{ range .Site.Params.custom_css }}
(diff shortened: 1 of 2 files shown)


budparr:
Thank you, @anthonyfok I'll try to go through this tonight. I've been totally swamped this week.

anthonyfok:
You are very welcome @budparr!  No hurries, nothing urgent, whenever you have time.

On second thought, relURL would work just as well absURL, and would make the generated HTML files slightly smaller, but that is entirely your preference, of course.  :-)

budparr:
I typically use relURL. Not sure why I would have done differently here.

anthonyfok:
> I typically use relURL. Not sure why I would have done differently here.

It might be due to @anddongyangli's suggestion in #97 where he said: "We might want to set **absolute URLs** for these assets based on the configured `baseURL`".

I have changed the `absURL` to `relURL` and pushed a new commit to this pull request for your convenience.  :-)  No rush, whenever you have free time.  Cheers!

anthonyfok:
Actually, forget my "ping" for now, especially if you are busy.

Now I am suspecting that Hugo's absURL is buggy because its current behaviour differs from that in Hugo 0.14.  In the current release, relURL is fine with a leading slash, but absURL is not, and that is inconsistent.  Hmm...  Need more investigation.

budparr:
Thank you @anthonyfok I'm actually fine with the change, I just haven't had time to run it myself, which I try to do. But I'm going to merge this since that's broken. Thanks for keeping on me about this.

PR #107, opened 2018-05-15
Add inheritance for social links.

LimitedEdition1:
This allows for social links to be either inherited from config.toml, specific for each page, or completely disabled per page.

### DIFF
--- a/layouts/partials/social-follow.html
+++ b/layouts/partials/social-follow.html
@@ -1,31 +1,31 @@
 <!-- TODO: Add follow intents where available TODO: Revisit color and hover color -->
 {{ $icon_size := "32px" }}
-{{ with .Site.Params.facebook }}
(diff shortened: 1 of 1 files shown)


budparr:
I like this idea, @LimitedEdition1 Thank you. I'm wondering if we need the global scope on the param here, `$.Param "twitter"` instead of `.Param "twitter"`

LimitedEdition1:
I have it working on one of my sites with .Param "key" and seems to do what I expected.  I could test the other way too.

LimitedEdition1:
I just did a quick test and they both seem to do the same thing.  A live Netlify site works without the $, but 
perhaps the $.Param "key" is a preferred method.

budparr:
I think it's fine. I suppose it only needs that scope when it's inside of the context of something else.

PR #120, opened 2018-08-08
Add I18n and introduce ".Site.Params.Dateformat"

vonloxley:
Translations are in i18n/*.toml.
If ".Site.Params.Dateformat" doesn’t exist, it falls back to the default.

### DIFF
--- a/i18n/de.toml
+++ b/i18n/de.toml
@@ -1,0 +1,20 @@
+[Recent]
+other = "Letzte"
+
+[All]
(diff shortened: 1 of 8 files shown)


budparr:
Hey, @vonloxley Thanks for this. We definitely need some i18n love in the theme, though I think it would be better to just have the skeleton or framework for it in this theme, rather than things specific to your use-case.

vonloxley:
No Problem.

PR #132, opened 2018-09-19
Use relative url function for custom CSS files

daniele-athome:
I don't know if it's the right way to address this issue, but it works in both server mode and publishing mode.


budparr:
This looks good, Thanks @daniele-athome

PR #133, opened 2018-09-28
Removed hardcoded theme sample hero image.

cdeguise:
This will allow the user to "blank" out the hero default set in the config. The if statement for blank was unreachable.

No changed needed to example site as the featured image is correctly specified in the front matter.


budparr:
Thank you, @cdeguise. Appreciate your patience.

PR #134, opened 2018-10-13
Added reading time and word count in article

looer:
At the beginning of the article, below the title, next to the pusblished date, are added both an _indication of the reading time_ and of _number of words_ contained in the article. See this example:

screenshot

It's implemented using the built-in Hugo **.ReadingTime** and **.WordCount** page variables.

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -28,6 +28,17 @@
     <main class="nested-copy-line-height lh-copy serif f4 nested-links nested-img mid-gray pr4-l w-two-thirds-l">
         {{- .Content -}}
       {{- partial "tags.html" . -}}
+
+      <div class="mt6">
+        {{ with .PrevInSection }}
+          << <a href="{{ .Permalink }}">{{ .Title }}</a>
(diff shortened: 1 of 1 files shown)


budparr:
Thank you, @looer This looks good. I'll review this weekend (sorry to be so slow!).

budparr:
Thank you for your work on this @looer I added it separately because I added some conditionals so that people who don't want this don't get it, but it's easy to set either globally, page-level, or at the section level.

```
Show "reading time" and "word count" but only if one of the following are true:
 1) A global config `params` value is set `show_reading_time = true`
 2) A section front matter value is set `show_reading_time = true`
 3) A page front matter value is set `show_reading_time = true`
  
   {{ if (or (eq (.Param "show_reading_time") true) (eq $section.Params.show_reading_time true) )}}
        <span class="f6 mv4 dib tracked"> - {{ .ReadingTime}} minutes read</span>
        <span class="f6 mv4 dib tracked"> - {{ .WordCount}} words</span>
      {{ end }}
```

PR #140, opened 2018-11-13
Make cover dimming class customisable.

andreyromancev:
60% dimming does not look great on all images. The ability to change that would be awesome!


budparr:
Thanks, @andreyromancev!

PR #141, opened 2018-11-15
Ensure description param is used.

ajduncan:
I'm a bit new to hugo and the theme, but I found to use the exampleSite config.toml's description parameter correctly, I had to update the site-header.html to use {{ with .Param "description" }} instead of {{ with .Param.description }}.  Sorry if this is redundant or wrong!

### DIFF
--- a/layouts/partials/site-header.html
+++ b/layouts/partials/site-header.html
@@ -9,7 +9,7 @@
         <h1 class="f2 f-subheadline-l fw2 white-90 mb0 lh-title">
           {{ .Title | default .Site.Title }}
         </h1>
-        {{ with .Params.description }}
+        {{ with .Param "description" }}
           <h2 class="fw1 f5 f3-l white-80 measure-wide-l center mt3">
             {{ . }}
           </h2>


robert-mcdermott:
This fixes the problem for me too please merge.

budparr:
Hey, @robert-mcdermott and @ajduncan This is a good addition, but I haven't merged it because it's not as small of a change as you might expect. Many users would suddenly have their globally set description appear on internal pages without expecting so. I'm going to keep this open and determine if we want to add it to a breaking-change type of release, but in the meantime, you can just duplicate the template in your layouts folder and make the change on your site.

immanuelpotter:
@ajduncan this worked for me too, cheers for that!

trevbhatt:
This fixed the issue for me too. I couldn't get the description to show up without it--am I missing something?

regisphilibert:
Thanks for this @ajduncan!

But @budparr is right. Users who don't want any description under their title on pages where a `.Params.description` is not set would have no other choice but to set the .Site.Params.description` to `false` which of course we don't want.

This would require the introduction of a new config key for this and even then we might introduce some breaking change. @ajduncan would you mind creating an issue to track this? I'll close this PR.

Thanks again 🙏

PR #142, opened 2018-11-17
Add custom head partial to basic template.

andreyromancev:
Hi!
Just added custom head partial to the basic template. Handy for importing fonts etc. Correct me if there is another way to do it please :)


budparr:
Hey, @andreyromancev Thank you for this. Some of these changes look pretty specific to your use-case and should just be overridden at your site's root instead of being changed in the theme. Closing for now, but let me know if you disagree or have questions about how to do that.

PR #146, opened 2018-12-20
Add explicit path to image example

caroso1222:
I'm new to Hugo and it wasn't straightforward to find the right file to see the image example.


budparr:
Thanks, @caroso1222!

Issue #145, opened 2018-12-19
SOcial Media links are not accessible


zersiax:
Hi,

On your theme's description page it lists that you aim to be accessible. However, the links for social media do not have textualcontent, making screen readers unable to determine where they go. ARIA-describeby is used to indicate the link opens in a new window, but the actual location of the link is unknown.
I have fixed it for now by adding text to the a tags in social-follow.html, but am fully blind myself so I have no idea to what degree that borks up the design. Can this be looked at?

zersiax:
I would keep the title of opens in a new window, and use aria-label for the actual function of the link. I think that will work best. If not, you might need to aria-hidden=true the icon itself, but I think just aria-label should suffice. Shoot me a test version and I'll compare.

PR #143, opened 2018-11-26
Open social media links in new tab and add Medium icon

hannahhenderson:
**Changes**
- [x] Opens social media links in a new tab
- [x] Adds an svg file of the Medium logo ([source](https://medium.design/logos-and-brand-guidelines-f1a01a733592))
- [x] Updates the assets

**Considerations**
1. Given that the changes are generated, I'm not sure how you'd like to see the `static/dist` files updated. Lmk if you'd like to see any changes.
2. I used the same `:hover` color as for LinkedIn.

medium-icon

**UPDATES**
Updated to reflect @budparr's input. These new changes include the addition of:
- [x] id span letting folks know when a link opens in a new window
- [x] `new-window` svg

I reduced the right-margin between social media icons to reduce the spread introduced by the hidden open-in-new-tab `new-window` svg.

screen shot 2018-11-27 at 11 17 39 am

with-accessibility-icon

PR #147, opened 2018-12-26
Improve social link accessibility

budparr:
Keep in spirit with https://medium.com/@svinkle/why-let-someone-know-when-a-link-opens-a-new-window-8699d20ed3b1 but address issue #145


budparr:
n.b. a continuation of the conversation in #143

PR #148, opened 2018-12-28
If nodate is set do not show the date line.

emin63:
For pages like about or bio or things like that, people may not want to have a date line.

With this change, you can have `nodate = true` set for a page to suppress the date.


emin63:
Thanks for telling me about the page feature. That solves my issue perfectly.

PR #154, opened 2019-02-03
Enable localization/modification of "Recent" string

naile:
Thanks for a great theme!

Super small PR for a thing i had to modify when attempting localization. Not sure of the name but it's in line with the `read_more_copy` in [summary-with-image.html](https://github.com/budparr/gohugo-theme-ananke/blob/85d9218877aea3f0f35fe7a887f714e09c2bc4fe/layouts/partials/summary-with-image.html#L23)


budparr:
Thanks, @naile

PR #155, opened 2019-02-10
Support hiding the featured image header text.

wallaceicy06:
I updated the documentation as well as changed the contact page in the
exampleSite to use this feature as a demo.


budparr:
Thanks, @wallaceicy06 !

Issue #108, opened 2018-05-23
Multilingual support


boydkelly:
Will this be coming one day?  It would be awesome addition.  Thanks.

jan-xyz:
I would be interested as well, I'm looking in to this right now. I like that Ananke is really basic and not so bloated like the other themes especially the ones supporting multilingual content. All it needs for me right now is provide a language picker in the menu or side-bar.

I'm trying to figure out how themes and templates work. If you have good resources to look things up, I'd be happy to write a PR. The official documentation is really basic and doesn't really help.

jan-xyz:
I created a [fork](https://github.com/jan-xyz/gohugo-theme-ananke) of your repository and started looking deeper into the theme stuff and it worked :) I also have to row back on my statement that the documentation doesn't help. I mostly copied stuff from the documentation and adjusted it to fit the overall appearance of the theme. It's still WIP but I think it can be helpful for others looking into it. I am also open to other suggestions and would be happy to get this into the main theme!

jan-xyz:
Partial multi language support was merged into master with #144.

@budparr what is your opinion regarding localization of strings? I could implement something like [this](https://gohugo.io/content-management/multilingual/#translation-of-strings) with a PR. I do think, however, that it will be an ongoing effort with adding more localizations in the future.

budparr:
@jan-xyz I'd like to make this a truly multi-lingual theme, so your contribution would be welcome. 

With a little bit of care, it can be multi-lingual without upsetting those who come into the theme with one language. I will chat more about that with @regisphilibert

jan-xyz:
Sounds good. I would suggest to aim for a solution that looks the same as it is right now as a default and everything else is an add-on.

jan-xyz:
@budparr, I think this can be closed now?

PR #156, opened 2019-02-19
add i18n translation support

jan-xyz:
With this Draft PR I want to get early feedback on my approach for tackling #108 and I would really appreciate some feedback from @budparr, @regisphilibert and anyone else involved.

I implemented a few strings that are easily visible but that should also include most of the strings. This theme is actually quiet and there isn't much text to translate in the first place I think.

I switched to this branch for my own blog and you can see the changes live at: https://blog.jan-steinke.de

So far I only have the translations for "de"(German) and "en"(English). The multi language support build into Hugo should default to "en" if I can trust [the documentation](https://gohugo.io/content-management/multilingual/#configure-languages), I haven't actually checked if this is true but it should make sure that anyone using this theme with single language should be fine as well.
> `defaultContentLanguage` sets the project’s default language. If not set, the default language will be `en`.

One thing I consciously removed is the support for `recent_copy` and `read_more_copy` as I understood that these are for multi language support and should be unnecessary if we build into i18n support.

Things need clarification before WIP is done:

- [x] removal of `..._copy` parameters
 - keep it for backwards compatibility (https://github.com/budparr/gohugo-theme-ananke/pull/156#issuecomment-465339954)
- [ ] which languages should be supported with the first implementation
- [ ] anything else?

### DIFF
--- a/i18n/de.toml
+++ b/i18n/de.toml
@@ -1,0 +1,17 @@
+[more]
+other = "Mehr"
+
+[allTitle]
+other = "Alle {{.Title }}"
+
+[recentTitle]
+other = "Neuste {{.Title }}"
+
+[readMore]
+other = "weiterlesen"
+
+[whatsInThis]
+other = "Was ist in dieser {{ .Type }}"
+
+[related]
+other = "Ähnliches"
--- a/i18n/en.toml
+++ b/i18n/en.toml
@@ -1,0 +1,17 @@
+[more]
+other = "More"
+
+[allTitle]
+other = "All {{.Title }}"
+
+[recentTitle]
+other = "Recent {{.Title }}"
+
+[readMore]
+other = "read more"
+
+[whatsInThis]
+other = "What's in this {{ .Type }}"
+
+[related]
+other = "Related"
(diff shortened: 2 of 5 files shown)


echarp:
This looks awesome! :+1: 

You don't like using "T" instead of the longer "i18n" command? Shorter but not as clear?

I'll help with french if possible.

jan-xyz:
Thanks, any support is appreciated! If you want make a PR to my branch with French translations. I can also see if I can give you Push rights on my branch directly, if you want.

I saw the `T` option but don't really have a strong opinion on it, whatever works best for the project. Maybe something else to decide on now?

budparr:
Thanks, @jan-xyz! @regisphilibert has suggested we use this construction for the strings:

`{{$.Param "read_more_copy" | default (i18n "readMore") }}`

otherwise, this could be a breaking change for many.

jan-xyz:
Sounds good to me and I fixed it with your suggestion. Is there anything else that you would like to see with this?

budparr:
Looks good to me. I'm going to run it locally before merging, but will try to get to that tonight. Thanks @jan-xyz!

jan-xyz:
No need to thank me :) I'm benefiting as well from this project.

PR #168, opened 2019-03-26
Fix TOC header

kwk:
In English, the translation that is supposed to be shown on top of the table of contents looks like this:

```toml
[whatsInThis]
other = "What's in this {{.Type }}"
```

When it is invoked, I think that something is wrong in the calling code: `{{ i18n "whatsInThis" humanize .Type }}`.

This should be called `{{ i18n "whatsInThis" . }` instead because then the i18n can access the `.Type`; otherwise you get a `<no value>` output.


budparr:
Thanks!

PR #170, opened 2019-04-30
Create zh.toml

lagagain:
Hi,
I translate a little part to Chinese.

Only one I don't know where it would be shown at.


budparr:
Thanks!

PR #172, opened 2019-05-09
Added a 'read more' button to the summary without image.

johanv:
I noticed that on the [front page](https://gohugo-ananke-theme-demo.netlify.com/), in the list with recent artciles, the summaries have a 'read more'-button. But if you navigate to the ['articles' page](https://gohugo-ananke-theme-demo.netlify.com/post/), there are no 'read more'-buttons in the summaries.
I'm not sure this is a bug or a feature.
But anyway, this merge request will add the 'read more' buttons where they are missing.

### DIFF
--- a/layouts/partials/summary.html
+++ b/layouts/partials/summary.html
@@ -9,5 +9,6 @@
     <div class="nested-links f5 lh-copy nested-copy-line-height">
       {{ .Summary  }}
     </div>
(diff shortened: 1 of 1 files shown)


budparr:
Thank you @johanv It was an intentional omission, I'm reluctant to make visual changes where people have been using the theme for a while, but I will consider this next time I do a design update. You can of course, override the file yourself.

PR #178, opened 2019-06-08
Add commento

agrrh:
Added configurable option to use https://commento.io/ block.

Sorry, my editor auto-removed some trailing whitespaces.

### DIFF
--- a/layouts/partials/commento.html
+++ b/layouts/partials/commento.html
@@ -1,0 +1,2 @@
+<div id="commento"></div>
+<script src="https://cdn.commento.io/js/commento.js"></script>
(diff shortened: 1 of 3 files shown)


layouts/partials/commento.html
```diff
@@ -0,0 +1,2 @@
+<div id="commento"></div>
+<script src="https://cdn.commento.io/js/commento.js"></script>
```
budparr: @agrrh Can you please add `defer` to the script as it specifies in the documentation? https://docs.commento.io/installation/cloud-service/

agrrh: Done!

budparr:
Thanks!

PR #179, opened 2019-06-17
Add portuguese translation

eriicf:
Portuguese translation. I've just updated by comparing to the English en.toml and Spanish es.toml.


budparr:
Thanks!

PR #183, opened 2019-07-10
Use git submodules

ndastur:
As per https://gohugo.io/getting-started/quick-start/ I think it is better to use git submodules rather than cloning the repo. It should be easier to keep up to date etc. So please consider this change in documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ This theme uses the "Tachyons" CSS library. This will allow you to manipulate th
 Inside the folder of your Hugo site run:
 
     $ cd themes
-    $ git clone https://github.com/budparr/gohugo-theme-ananke.git
+    $ git submodule add https://github.com/budparr/gohugo-theme-ananke.git themes/ananke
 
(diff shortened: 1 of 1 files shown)


budparr:
Thank you. I think a lot of users will have a difficult time with Git submodules. I'm happy to add this as an alternative, but not in replace of the cloning method.

ndastur:
Agreed git sub-modules are a PITA but it's the officially documented way to deal with themes in Hugo and for the initial clone no real difference. The difference comes when you want to update from the repo. At that point "users" who have done a git clone essentially get stuck. As it is difficult to update the code into their own site's repo.

Can I make a suggestion that if you keep the git clone rather than the submodule. Could you please add instructions on how to keep code up to date from your repository after cloning and creating a user's own site.

BTW great theme and thanks for putting the theme out there under MIT.

budparr:
Thanks. Actually, I'm going to go with your change. I may add some language to it later though.

PR #185, opened 2019-07-15
localization for form-contact.html shortcode

### LINKED ISSUE #184: text in form-contact.html is not localized
The text of the [contact-form](https://github.com/budparr/gohugo-theme-ananke/blob/master/layouts/shortcodes/form-contact.html) shortcode is not translated. Should be added to i18n.

eddex:
I translated the static text in the form-contact.html shortcode to all supported langages.

German is my mother tongue, the rest I translated with [DeepL](https://www.deepl.com/translator) (except for Chinese, there I had to use Google Translate).

This fixes #184

### DIFF
--- a/i18n/de.toml
+++ b/i18n/de.toml
@@ -15,3 +15,18 @@ other = "Was ist in dieser {{ .Type }}"
 
 [related]
 other = "Ähnliches"
+
+[yourName]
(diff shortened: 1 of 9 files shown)


budparr:
Thanks, @eddex

PR #187, opened 2019-07-26
fix parse error in theme.toml

rburgst:
Otherwise I am getting
```
Error: failed to read module config for "ananke" in "/builds/rburgst/focusgtdapp-web/themes/ananke/theme.toml": unmarshal failed: Near line 11 (last key parsed 'min_version'): Invalid float value: "0.30.2"
```

### DIFF
--- a/theme.toml
+++ b/theme.toml
@@ -8,7 +8,7 @@ description = "A Base theme for building full featured Hugo sites"
 homepage = "https://github.com/budparr/gohugo-theme-ananke"
 tags = ["website", "starter", "responsive", "Disqus", "blog", "Tachyons", "Multilingual"]
 features = ["posts", "shortcodes", "related content", "comments"]
-min_version = 0.31
+min_version = "0.30.2"
 
 [author]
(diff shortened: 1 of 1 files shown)


theme.toml
```diff
@@ -8,7 +8,7 @@ description = "A Base theme for building full featured Hugo sites"
 homepage = "https://github.com/budparr/gohugo-theme-ananke"
 tags = ["website", "starter", "responsive", "Disqus", "blog", "Tachyons", "Multilingual"]
 features = ["posts", "shortcodes", "related content", "comments"]
-min_version = 0.30.2
+min_version = 0.31
```
jasmith86: Per the conversation in the referenced issue, it may be better to change `min_version = 0.30.2` to a string instead:

`min_version = "0.30.2"`

Thoughts?

rburgst: works, too, see updated PR

aknuds1:
Experiencing the same issue!

budparr:
Closing as fixed in #189

Issue #186, opened 2019-07-25
Prolem parsing theme.toml


jasmith86:
When attempting to make add a new post with `hugo new posts/my-post.md`, I get the following error:

```
Error: failed to read module config for "ananke" in "/home/user/Documents/projects/myhugosite/themes/ananke/theme.toml": unmarshal failed: Near line 11 (last key parsed 'min_version'): Invalid float value: "0.30.2"
```

Editing `theme.toml` and commenting out line 15 `min_version = 0.30.2` allows the hugo command to execute properly.

rburgst:
same problem here, I guess newer hugo versions parse this version more strict.

rburgst:
I have created a PR for this #187

amir3code:
@rburgst  I think thats because of theme.toml file being parsed. I mean the parser has problem with that. I changed that to string `min_version = "0.30.2"` that would not be a problem right?

rburgst:
Most likely yes. I am not sure whether it makes a huge difference (after all 0.30 is already pretty old). I have my build working with my PR but I guess it would work with a string, too.
I checked a few online resources and all of them list `min_version` as a float.

theHausdorffMetric:
I think one should use a string "0.30.2" and not a float to be consistent with hugo's andard 3-part semantic versioning scheme. 
See also [https://github.com/gohugoio/hugo/issues/4127](url)

rburgst:
I have updated the PR and it also works. So I would say lets go with string then.

budparr:
This has been fixed in #189

PR #188, opened 2019-08-02
Update theme.toml

heyestom:
I was getting this error using while flowing the hugo quick start, changing to a string fixed it. 
```
Error: failed to read module config for "ananke" in "/Users/theyes/Projects/hugo/demo/themes/ananke/theme.toml": unmarshal failed: Near line 11 (last key parsed 'min_version'): Invalid float value: "0.30.2"
```


heyestom:
#186 is the same, I'll close this.

PR #189, opened 2019-08-02
Fix min_version

bep:

PR #192, opened 2019-08-05
fix toml format

Extrawurst:
without `Hugo` complained with:
```
Error: failed to read module config for "ananke" in "[...]/quickstart/themes/ananke/theme.toml": unmarshal failed: Near line 11 (last key parsed 'min_version'): Invalid float value: "0.30.2"
```


budparr:
Fixed in #189

PR #199, opened 2019-08-18
Update site-footer.html

jorgeml:
Change footer to allow changing the copyright as a site parameter.

I'm not an HTML expert so I'm not sure about the formatting.

### DIFF
--- a/layouts/partials/site-footer.html
+++ b/layouts/partials/site-footer.html
@@ -1,8 +1,8 @@
 <footer class="{{ .Site.Params.background_color_class | default "bg-black" }} bottom-0 w-100 pa3" role="contentinfo">
   <div class="flex justify-between">
-  <a class="f4 fw4 hover-white no-underline white-70 dn dib-ns pv2 ph3" href="{{ .Site.BaseURL }}" >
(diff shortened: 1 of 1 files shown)


budparr:
Thank you for this. I think we should use Hugo's built in Site Config value and use Hugo's default function instead of `else`. For example:
```
{{ with .Site.Copyright | default "&copy;" }} {{ . | safeHTML }} {{ now.Format "2006"}} {{ end }}
```

jorgeml:
> Thank you for this. I think we should use Hugo's built in Site Config value and use Hugo's default function instead of `else`. For example:
> 
> ```
> {{ with .Site.Copyright | default "&copy;" }} {{ . | safeHTML }} {{ now.Format "2006"}} {{ end }}
> ```

Fine by me, I'm far from being a Hugo expert and if that works...

budparr:
Do you want to make the change in your PR, or shall I do it myself?

jorgeml:
Would you mind doing the change yourself please? I don't want to break anything.

budparr:
No problem at all. Thanks for bringing it up.

PR #202, opened 2019-08-21
Use environment introduced in Hugo 0.53

### LINKED ISSUE #201: robots.txt disallows everything despite "Environment: production"
I wanted to add `robots.txt` to my site so I added

```
enableRobotsTXT: true
```

to my `config.yaml`. That didn't turn out as expected. It got the following content:

```
User-agent: *
# robotstxt.org - if ENV production variable is false robots will be disallowed.

  Disallow: /

```

On the web server I render my site using just `hugo`. On my workstation I tried with `hugo server --environment=production` but got the same result. The command wrote the environment as I specified it however:

```
$ hugo server --environment=production
...
Environment: "production"
Serving pages from memory
...
```

I'm running Hugo 0.57.2.

After some digging I found that `layouts/robots.txt` in this git is looking for the environment variable `HUGO_ENV` or the configuration variable `params.env` (`Site.Params.env`).

My conclusion is that the Ananke theme got environment support before Hugo did and when Hugo added it in [version 0.53](https://gohugo.io/news/0.53-relnotes/) it was in a slightly different way. Running `hugo server` will set the environment to _development_ and running `hugo` will set it to _production_. It can be overridden by OS environment variable `HUGO_ENVIRONMENT`(note, not `HUGO_ENV`!) or by using the `--environment` option.

My opinion is that the Ananke theme shall support the official way of specifying environment.

labanskoller:
The Ananke theme got environment support before Hugo did and when Hugo
added it in version 0.53 it was in a slightly different way. Running
"hugo server" will set the environment to "development" and running just
"hugo" will set it to "production". It can be overridden by OS
environment variable "HUGO_ENVIRONMENT" (note, not "HUGO_ENV") or by
using the --environment option.

Use "hugo.Environment" to detect environment instead and update the
README.

Resolves #201.

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -8,7 +8,7 @@
     <meta name="viewport" content="width=device-width,minimum-scale=1">
     {{ .Hugo.Generator }}
     {{/* NOTE: For Production make sure you add `HUGO_ENVIRONMENT="production"` before your build command */}}
-    {{ if eq hugo.Environment "production" }}
+    {{ if eq (getenv "HUGO_ENV") "production" | or (eq hugo.Environment "production") }}
       <META NAME="ROBOTS" CONTENT="INDEX, FOLLOW">
     {{ else }}
       <META NAME="ROBOTS" CONTENT="NOINDEX, NOFOLLOW">
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -6,9 +6,9 @@
     {{/* NOTE: the Site's title, and if there is a page title, that is set too */}}
     <title>{{ block "title" . }}{{ .Site.Title }} {{ with .Params.Title }} | {{ . }}{{ end }}{{ end }}</title>
     <meta name="viewport" content="width=device-width,minimum-scale=1">
-    {{ hugo.Generator }}
-    {{/* NOTE: For Production make sure you add `HUGO_ENV="production"` before your build command */}}
-    {{ if eq (getenv "HUGO_ENV") "production" | or (eq .Site.Params.env "production")  }}
+    {{ .Hugo.Generator }}
+    {{/* NOTE: For Production make sure you add `HUGO_ENVIRONMENT="production"` before your build command */}}
+    {{ if eq hugo.Environment "production" }}
       <META NAME="ROBOTS" CONTENT="INDEX, FOLLOW">
     {{ else }}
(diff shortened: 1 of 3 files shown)


budparr:
This still replaces `HUGO_ENV` with `HUGO_ENVIRONMENT` in the readme and I think we need to leave the other config param in until people can be notified to some degree.

labanskoller:
Sorry for the force push. I guess updating feature branches the GitHub way is by merging.
I've now reverted all my README changes. Do you think this can go upstream now?

layouts/_default/baseof.html
```diff
@@ -6,9 +6,9 @@
     {{/* NOTE: the Site's title, and if there is a page title, that is set too */}}
     <title>{{ block "title" . }}{{ .Site.Title }} {{ with .Params.Title }} | {{ . }}{{ end }}{{ end }}</title>
     <meta name="viewport" content="width=device-width,minimum-scale=1">
-    {{ hugo.Generator }}
-    {{/* NOTE: For Production make sure you add `HUGO_ENV="production"` before your build command */}}
-    {{ if eq (getenv "HUGO_ENV") "production" | or (eq .Site.Params.env "production")  }}
+    {{ .Hugo.Generator }}
```
regisphilibert: We should stick with `hugo`'s global function here.

layouts/_default/baseof.html
```diff
@@ -39,12 +39,12 @@
     {{- template "_internal/schema.html" . -}}
     {{- template "_internal/twitter_cards.html" . -}}
 
-    {{ if eq (getenv "HUGO_ENV") "production" | or (eq .Site.Params.env "production")  }}
+    {{ if eq hugo.Environment "production" }}
```
regisphilibert: Any reason we're not testing for both `HUGO_ENV` or `hugo.Environement` here like on line 11?

budparr:
We need to do this in a manner that wouldn't break sites. We can't do breaking changes without advance warning and this is not a big enough change to warrant announcements and such.

regisphilibert:
I just realized one thing... From now on any deployement using the simple `hugo` command will pass the production test. This might be a breaking change as people using `HUGO_ENV` set to `staging` will have their build test positive.

labanskoller:
I've realized that what I thought was easy was in fact not for me. I don't know enough of Hugo or how it's used by some sites and I don't come here often enough to collaborate on this pull request. It might also be that something is lost in translation. I could have misunderstood something basic.

I hereby close this pull request and hope that somebody else resolves issue #201 in the future.

PR #204, opened 2019-09-10
Fix README instructions

gilbsgilbs:
`themes` directory is already set in the submodule command. Therefore, you should not set it as working directory.


budparr:
thank you.

Issue #113, opened 2018-06-15
Adding Author to Blog Posts


MayaGans:
Hi! I was wondering if it'd be possible to add the blog post author to the top of each post page, possibly automating this? [I found a blog post related to this](https://discourse.gohugo.io/t/how-to-access-site-params-map-with-a-params-variable/348), but because I'm a total n00b I'm not sure how to go about implementing it within the theme or where to add this script within the css. Any tips helpful, and thanks for the wonderful blog layout!

budparr:
Hi @MayaGans You'd want to duplicate this template: https://github.com/budparr/gohugo-theme-ananke/blob/master/layouts/_default/single.html into your root layouts directory (so it would be like `/layouts/_default/single.html` instead of `/gohugo-theme-ananke/blob/master/layouts/_default/single.html` and add the code there. Unfortunately, I don't have time to help you with the actual code, but it's definitely doable. People in the forums would be able to help you with that and I'll consider adding authors in a future version of the theme.

Beanow:
I've added a PR above. It's a simple implementation that doesn't go all over the place like lists to add it. However it's there on the blog post itself.

For the website I'm writing for now I believe it's necessary, because part of the site is written from "we" perspective. But blog posts can also be opinion written in "I" form which needs the author for this distinction.

PR #209, opened 2019-09-17
Adds an author to blog posts.

Beanow:
See #113
This adds an author to the blog page only.
Implementation mostly based on Docsy's version.
https://github.com/google/docsy/blob/master/layouts/blog/content.html

image

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -18,6 +18,9 @@
       <h1 class="f1 athelas mb1">
         {{- .Title -}}
       </h1>
+      <p class="tracked">
+        {{ with .Params.author }}By <b>{{ . | markdownify }}</b>{{ end}}
+      </p>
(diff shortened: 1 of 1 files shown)


budparr:
Hey, @Beanow I'm basically fine with this. Using `author` could be problematic, because some may use `author` for something like a taxonomy lookup, but I think it should be okay.

However, could you wrap the entire p tag in the `with` statement?

So, instead of 
```
<p class="tracked">
    {{ with .Params.author }}By <b>{{ . | markdownify }}</b>{{ end}}
</p>
```
use:
```
{{ with .Params.author }}
    <p class="tracked">
        By <strong>{{ . | markdownify }}</strong>
     </p>
{{ end}}
```

Beanow:
Would have been happy with you making the change, but here you go :+1:

PR #210, opened 2019-09-20
html, json, toml and markdown files don't require executable permission

junjieyuan:


budparr:
Hi @JunjieYuan I'm not sure I understand this PR. Could you explain, please?

JonathanDoughty:
On Linux & MacOS at least the referenced files end up with permissions like
`-rwxr-xr-x`
There is no need for what are essentially text files to have the x (executable) permission so the PR removes those, making them
`-rw-r--r--`
It is a minor annoyance, sometimes exacerbated by terminals that colorize files differently depending on the permissions. 

This most likely comes from something that originated on a Windows filesystem which does not have the same permission scheme. For more on this wrt git see https://git-scm.com/docs/git-config#Documentation/git-config.txt-corefileMode and/or this SO question & responses that details some of the pain this can cause https://stackoverflow.com/questions/1580596/how-do-i-make-git-ignore-file-mode-chmod-changes

budparr:
Hi folks. Can someone explain to me why this needs to be in the public version of the theme?

JunjieYuan:
This PR may not improve security, but at least it won't make it worse :)

garrison:
I was about to make a pull request myself removing the executable bit from these files.  Having the executable bit set is a special marker, and it is confusing to have it set for files that do not need it.

PR #211, opened 2019-09-26
init

houcl:


budparr:
H @houcl Could you explain to me what's going on in the footer? It looks like all users would have those characters, no?

And perhaps tell me what the intent with the slug is as well.

Thank you!

junjieyuan:
> H @houcl Could you explain to me what's going on in the footer? It looks like all users would have those characters, no?
> 
> And perhaps tell me what the intent with the slug is as well.
> 
> Thank you!

The content of this PR appears to be directed only to his own website.

He changed the translation from (Traditional) Chinese to Simplified Chinese. The information in the footer is the record information of the website in mainland China. If you want to open a website in mainland China, you need to record it (Bei An).

PR #216, opened 2019-10-13
Replace {{ .URL }} with {{ .Permalink }}

jish:
Using the `{{ .URL }}` token results in a warning:

    WARN 2019/10/13 15:12:12 Page's .URL is deprecated and will be removed in a
    future release. Use .Permalink or .RelPermalink. If what you want is the
    front matter URL value, use .Params.url.

### Steps to reproduce

Follow along with the Quick Start guide at
https://gohugo.io/getting-started/quick-start/

Running `hugo server -D` emits a warning that `.URL` is deprecated.


budparr:
Thanks!

PR #219, opened 2019-10-26
Fix compile error caused by MenuEntry not having a .Permalink

N0hbdy:
Fixes issue in recent change that changes {{.URL}} to {{.Permalink}},
but shouldn't have been done for navigation.MenuItems.

To reproduce: `hugo server -D` will spew errors, so long as you have a menu defined in config.toml

After updating to this fix, hugo resumes building normally.


budparr:
Fixed in #230 Thanks!

PR #221, opened 2019-10-27
Fix to take care of multiple author list

frankieliu:
In case the author parameter is a []string, which is the default for ox-hugo
output.

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -20,7 +20,13 @@
       </h1>
       {{ with .Params.author }}
       <p class="tracked">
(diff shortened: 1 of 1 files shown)


budparr:
Thanks for this, @frankieliu I've asked @regisphilibert to have a second set of eyes when he gets a chance.

tompscanlan:
This looks good to me.  Adding:

    author:
      - "Tom Scanlan"
      - "Foo Bar"

shows:  
    By **Tom Scanlan, Foo Bar**

budparr:
Thank you, @tompscanlan

PR #222, opened 2019-10-27
Fix word count heading typo in README.md

mnihalm:
The heading says "Word Contributing"  where it should say "Word Count"


budparr:
Thanks, @mnihalm

PR #236, opened 2019-11-09
Add Slack to social options

jacebrowning:
I downloaded and modified the SVG from https://brandfolder.com/slack/logos. It's still a bit too small, but couldn't find a better alternative.


budparr:
Thanks!

PR #237, opened 2019-11-27
Fix Slack icon size

jacebrowning:
The hover color is still not working, but now the icon size matches the others:


jacebrowning:
@budparr Any chance we could get this merged? It will really improve the look of our site. 😄

budparr:
Thanks!

PR #240, opened 2019-12-27
Use Hugo's built in Site Config for copyright according to PR #199

giacomd:
Hello and thanks for this amazing theme!
It looks like the modifications discussed at #199 didn't go through, so here they are again.
Cheers


budparr:
Thanks!

z3ntu:
Is it now `(c) $sitename $year` instead of `(c) $year $sitename` on purpose?

budparr:
That's because sitename is the default for when you don't have copyright info in your config.

PR #241, opened 2020-01-01
Correct the original translation

zhangsheng377:
Some original translations are not universal Simplified Chinese.


budparr:
Guess I'll just have to trust you on these :) 
Thanks!

Issue #81, opened 2018-02-04
Site logo


Tim-Butterfield:
It would be nice to include support for a site logo image parameter.  One possibility might be a modification to the BaseURL link in site-navigation.html like this:

```
    <a href="{{ .Site.BaseURL }}" class="f3 fw2 hover-white no-underline white-90 dib">
      {{ if .Site.Params.logo }}
      <img src="{{ .Site.Params.logo }}" class="v-mid" />
      {{ end }}
      {{ .Site.Title }}
    </a>
```

budparr:
I'm going to do some work on the theme this week and this shouldn't be too hard to add. Thanks, @Tim-Butterfield

kentmz:
I'd love to see this feature to add a logo.

PR #249, opened 2020-01-31
Add the option to use a logo for the top left site title

nickjwhite:
This fixes issue #81, but it does so slightly differently. It replaces the site title with a logo, in the navigation section, if Site.Params.Logo is set. If not it just continues to print the title, as it has up 'til now. Rather than Lithium's approach of having dimensions set under the logo parameter (see https://github.com/jrutheiser/hugo-lithium-theme/blob/master/exampleSite/config.toml) I just went for simplicity, figuring people would be happy to just set the dimensions in the image, or CSS, themselves.

This is my first hugo theme adjustment, so please check it over for any stupid mistakes!

### DIFF
--- a/layouts/partials/site-navigation.html
+++ b/layouts/partials/site-navigation.html
@@ -1,7 +1,11 @@
 <nav class="pv3 ph3 ph4-ns" role="navigation">
   <div class="flex-l justify-between items-center center">
     <a href="{{ .Site.BaseURL }}" class="f3 fw2 hover-white no-underline white-90 dib">
+    {{ if .Site.Params.Logo }}
+      <img src="{{ .Site.Params.Logo | relURL }}" alt="{{ .Site.Title }}" />
+    {{ else }}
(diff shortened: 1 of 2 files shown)


layouts/partials/site-navigation.html
```diff
@@ -1,7 +1,11 @@
 <nav class="pv3 ph3 ph4-ns" role="navigation">
   <div class="flex-l justify-between items-center center">
     <a href="{{ .Site.BaseURL }}" class="f3 fw2 hover-white no-underline white-90 dib">
+    {{ if .Site.Params.Logo }}
```
regisphilibert: Thanks for this. Using `with` would make this a tiny bit less verbose (still for best practices promoting):

```
{{ with .Site.Params.Logo }}
  <img src="{{ relURL . }}" alt="{{ $.Site.Title }}" />
{{ else }}
   {{ .Site.Title }}
{{ end }}

PR #250, opened 2020-01-31
Logo with hidpi support

nickjwhite:
An alternative to PR #249, this one includes the option to use a hidpi version of the logo, implementing it with srcset. Also addresses issue #81.


jtagcat:
This and #249 does not address sizing, adding `height="48pt"` (though this was eyeballing and could be improved) 
image

PR #251, opened 2020-01-31
Use relative URL for favicon

nickjwhite:
This fixes sites which have the baseURL in a subdirectory.


layouts/partials/site-favicon.html
```diff
@@ -1,3 +1,3 @@
 {{ if .Site.Params.favicon }}
-<link rel="shortcut icon" href="/{{ .Site.Params.favicon }}" type="image/x-icon" />
-{{ end }}
\ No newline at end of file
+<link rel="shortcut icon" href="{{ relURL ($.Site.Params.favicon) }}" type="image/x-icon" />
```
regisphilibert: Thanks again. And again could you do as for #252 and remove the () :)

Thanks a ton!

jan-xyz: May I make it easier to change? :)
```suggestion
<link rel="shortcut icon" href="{{ relURL $.Site.Params.favicon }}" type="image/x-icon" />
```

PR #252, opened 2020-01-31
Fix relURL for custom_css

nickjwhite:
The way relURL was used meant that one needed to preface the custom_css
entries with a slash, or the href would be incorrectly set to
subdircss/mycss.css rather than subdir/css/mycss.css, in the case of
a baseURL in a subdir.

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -20,7 +20,7 @@
     {{ end }}
 
     {{ range .Site.Params.custom_css }}
-      <link rel="stylesheet" href="{{ relURL ($.Site.BaseURL) }}{{ . }}">
+      <link rel="stylesheet" href="{{ relURL (.) }}">
     {{ end }}
 
     {{ block "favicon" . }}


layouts/_default/baseof.html
```diff
@@ -20,7 +20,7 @@
     {{ end }}
 
     {{ range .Site.Params.custom_css }}
-      <link rel="stylesheet" href="{{ relURL ($.Site.BaseURL) }}{{ . }}">
+      <link rel="stylesheet" href="{{ relURL (.) }}">
```
regisphilibert: @nickjwhite thank you for this.

Would you mind removing the () while you're at it. I understand they were there initially, but it turns out they're not needed.
```
<link rel="stylesheet" href="{{ relURL . }}">
```

This reads a bit picky, but we're getting back at improving the theme and use it to promote Hugo's best practices.

jan-xyz, line 23: For your convenience:
```suggestion
      <link rel="stylesheet" href="{{ relURL . }}">
```

I hope that makes it easier to fix!

regisphilibert:
@jan-xyz sorry for merging this so late! Thanks!

PR #255, opened 2020-02-05
Add sharing links for the posts

### LINKED ISSUE #198: Add sharing links for the posts
Medium has a way to share post to Twitter, Facebook

Screen Shot 2019-08-15 at 11 17 41 AM

It would be great to put such links in hugo blogs too. And we can also let people choose what platforms links they want to show and let them embed their usernames too in the post sharing the blog post. 

Some of the platforms / buttons that I have thought is - Twitter, Facebook, WhatsApp, Copy URL

 If this looks cool, I would like to raise a PR for this 😄

mountainbug95:
Fixes #198

### DIFF
--- a/layouts/partials/social-share.html
+++ b/layouts/partials/social-share.html
@@ -4,25 +4,19 @@
 {{ $body := print $title ", by " .Site.Title "\n" .Params.description "\n\n" $url "\n" }}
 {{ $icon_size := "32px" }}
 {{ $test_url := printf "%s" "https://www.google.com/" }}
-{{ if .Params.social_share.display_social_share }}
-  {{ $social := .Params.social_share.social_items }}
+
+{{ if not .Params.disable_share }}
   <div id="sharing" class="mt3">
-    {{ with $social.facebook }}
-      <a href="http://www.facebook.com/sharer.php?u={{ $url }}" class="facebook no-underline" aria-label="share on Facebook">
-        {{ partialCached "svg/facebook.svg" (dict "size" $icon_size) $icon_size }}
-      </a>
-    {{ end }}
+    <a href="http://www.facebook.com/sharer.php?u={{ $url }}" class="facebook no-underline" aria-label="share on Facebook">
+      {{ partialCached "svg/facebook.svg" (dict "size" $icon_size) $icon_size }}
+    </a>
 
-    {{ with $social.twitter }}
-      <a href="http://twitter.com/share?url={{ $url }}&text={{ $title }}&via={{with $site.Social.twitter }}{{ . }}{{ end }}" class="twitter no-underline" aria-label="share on Twitter">
-        {{ partialCached "svg/twitter.svg" (dict "size" $icon_size) $icon_size }}
-      </a>
-    {{ end }}
+    <a href="http://twitter.com/share?url={{ $url }}&text={{ $title }}&via={{with .Site.Social.twitter }}{{ . }}{{ end }}" class="twitter no-underline" aria-label="share on Twitter">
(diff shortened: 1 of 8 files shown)


regisphilibert:
Thanks for thinking about a config but two things:

1. We need a "disable" sharing config, as I believe by default, user will want those on and we can expect to turn on the setting on each post.
2. I don't think the user will want to limit the number of share services added to their post, for many: the more the better. But if we decide to allow this it should be on the project level. I don't see any reason why user would want Facebook on one post and not the other.

@budparr for 2. do you think we should allow user to disable a given share service? I don't but I might be missing some use cases.

budparr:
I agree. I don't think that fine-tuned disabling is necessary

layouts/partials/social-share.html
```diff
@@ -4,25 +4,19 @@
...
+    <a href="http://www.facebook.com/sharer.php?u={{ $url }}" class="facebook no-underline" aria-label="share on Facebook">
+      {{ partialCached "svg/facebook.svg" (dict "size" $icon_size) $icon_size }}
+    </a>
 
-    {{ with $social.twitter }}
-      <a href="http://twitter.com/share?url={{ $url }}&text={{ $title }}&via={{with $site.Social.twitter }}{{ . }}{{ end }}" class="twitter no-underline" aria-label="share on Twitter">
-        {{ partialCached "svg/twitter.svg" (dict "size" $icon_size) $icon_size }}
-      </a>
-    {{ end }}
+    <a href="http://twitter.com/share?url={{ $url }}&text={{ $title }}&via={{with .Site.Social.twitter }}{{ . }}{{ end }}" class="twitter no-underline" aria-label="share on Twitter">
```
regisphilibert: Sorry for not catching this earlier, but we should test fo `.Site.Social.Twitter` before adding the via parameter.

Might be better to first build your URL with printf for every service, and then use the variable as href. For twitter, this would allow you to first create a string without the via. Something like:
```
{{ $twitter_href := printf "https://twitter.com/share?url=%s&text=%s" $url $title }}
{{ with .Site.Social.twitter }}
  {{ $twitter_href = printf "%s&via=%s $twitter_href . }}
{{ end }}
```

Also, we should use `https` for those services. I was pondering using `//twitter.com` to ensure the protocole is relative and same as user's site but this might not be ideal if one is running `http` and twitter stop serving their service over it. So let's hardcode `https`

layouts/partials/social-share.html
```diff
@@ -0,0 +1,28 @@
+{{ $site :=  .Site }}
```
regisphilibert: `$site` is not used.

layouts/partials/social-share.html
```diff
@@ -0,0 +1,28 @@
...
+{{ $title := .Title }}
+{{ $url := printf "%s" .URL | absLangURL }}
+{{ $body := print $title ", by " .Site.Title "\n" .Params.description "\n\n" $url "\n" }}
+{{ $icon_size := "32px" }}
+{{ $test_url := printf "%s" "https://www.google.com/" }}
```
regisphilibert: `$test_url` is not used

layouts/partials/social-share.html
```diff
@@ -0,0 +1,28 @@
+{{ $site :=  .Site }}
+{{ $title := .Title }}
+{{ $url := printf "%s" .URL | absLangURL }}
+{{ $body := print $title ", by " .Site.Title "\n" .Params.description "\n\n" $url "\n" }}
```
regisphilibert: `$body` is not used

PR #259, opened 2020-02-17
 Long urls or links extend beyond content and overlap sidebar

### LINKED ISSUE #118: Long urls or links extend beyond content and overlap sidebar
If you have a really long url or link it will extend beyond the border for the main content instead of wrapping. If you have a sidebar it will overlap the sidebar content. As an example see the links at the bottom of the following page https://kitefaster.com/2017/05/22/create-today-widget-ios-app-swift/

It looks like the header and main tags are set to different widths in the CSS.

mountainbug95:
Fixes #118


mountainbug95:
@regisphilibert I've added a long link to the content of chapter-6.md. This can probably be removed before merging.

regisphilibert:
Thanks for this @mountainbug95 don't we have a way to only apply this rule to the `a` in the content of a page? This global rule might be problematic in the future.

PR #260, opened 2020-02-17
Font not applying to blog posts

### LINKED ISSUE #182: Font not applying to blog posts
When I try to change the global font in the config.toml file as described in the documentation I notice that it changes it for most of the site except not in the "posts" folder for the blog posts. I imagine I'm missing something but I don't see why the fonts used are different.

mountainbug95:
Fixes #182

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,8 @@ which will give you a body class like this:
 <body class="avenir bg-near-white">
 ```
 
+note: The `body_classes` parameter will not change the font used in post content. To do this, you must use the `post_content_classes` parameter.
+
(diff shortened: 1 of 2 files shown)


mountainbug95:
@budparr @regisphilibert I've added a separate parameter for changing the post content font. It looks like the theme was designed with most of the site content to be in a san-serif font but the post content to be in a serif font. I think it would be better to retain the ability to use a different font for post content than is being used for the rest of the site. What do you guys think?

budparr:
> I think it would be better to retain the ability to use a different font for post content than is being used for the rest of the site. What do you guys think?

Yes, I agree

PR #261, opened 2020-02-23
improved italian translation ([recentTitle] field)

fpiantini:
I started to use your theme for my personal renewed site, and I found a unsatisfactory translation for the `recentTitle` item in the translation file for the italian language.
With the current translation the italian resulting string is "Recenti posts" that does not sound right at all. A better translation is "Post recenti" but please note that Post should be used in singular (in italian, foreign terms are always used singular). I do not know if it is possibile to use `{{.Title }}` in some way to have it in singular form, so for the moment I decided to use another better (in my opinion) translation using the italian word "Articoli" instead of "Posts".
I am not sure but probably the same consideration apply to other translation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,12 @@ To run in production (e.g. to have Google Analytics show up), run `HUGO_ENV=prod
 HUGO_ENV=production hugo
 ```
 
+Note: The above command will not work on Windows. If you are running a Windows OS, use the below command:
+
+```
+hugo --environment production
+```
+
 ## Contributing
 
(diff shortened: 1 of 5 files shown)


regisphilibert:
@fpiantini thanks for this PR.

Your change would make the "name" of the content type hardcoded. We want it to dynamically pick up the name of the content type (Articoli, Ricette, etc... whatever user has in there).
I think we'll have to simply put `{{ .Title }}` at the beginning of the string providing we don't end up with gender typos in some cases... I don't speak italian, so you tell me.

In any case could you create a new PR with the change discussed above? 🙏 

I'm closing this PR as it has some unrelated commits including that 'mermaid' addition which should be part of another PR.

PR #275, opened 2020-05-11
Replace another 2 .URL occurrences with .Permalink

strk:
This is basically a completion of #216

### DIFF
--- a/layouts/partials/site-navigation.html
+++ b/layouts/partials/site-navigation.html
@@ -9,7 +9,7 @@
         <ul class="pl0 mr3">
           {{ range .Site.Menus.main }}
           <li class="list f5 f4-ns fw4 dib pr3">
-            <a class="hover-white no-underline white-90" href="{{ .URL }}" title="{{ .Name }} page">
+            <a class="hover-white no-underline white-90" href="{{ .Permalink }}" title="{{ .Name }} page">
               {{ .Name }}
             </a>
           </li>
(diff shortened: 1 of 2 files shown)


strk:
\cc @budparr @jish @bstoney

layouts/partials/site-navigation.html
```diff
@@ -9,7 +9,7 @@
         <ul class="pl0 mr3">
           {{ range .Site.Menus.main }}
           <li class="list f5 f4-ns fw4 dib pr3">
-            <a class="hover-white no-underline white-90" href="{{ .URL }}" title="{{ .Name }} page">
+            <a class="hover-white no-underline white-90" href="{{ .Permalink }}" title="{{ .Name }} page">
```
budparr, line 12: Did you test this? This one is incorrect. For menus, the `.URL` remains. https://gohugo.io/templates/menu-templates/

strk: I did not test, other than verifying that the following annoying message goes away with the patch:
> WARN 2020/05/11 16:14:04 Page's .URL is deprecated and will be removed in a future release. Use .Permalink or .RelPermalink. If what you want is the front matter URL value, use .Params.url.

I'm new to Hugo so don't really know how to test site-navigation (and I'm actually missing it, see http://strk.kbt.io/blog/)

budparr: No worries. We can just take this out. I'll have a look at the other one on my local, but should  be good.

budparr:
The second one is fine. If you could remove the one in the navigation, please, I'll merge this.

strk:
Done, and confirm the WARN does not re-surface

budparr:
Thanks!

PR #276, opened 2020-05-20
Updated Windows instructions in README.md

### LINKED ISSUE #153: Windows environment variables are not considered
This command in the README is invalid when using hugo on Windows:
```
HUGO_ENV=production hugo
```
I set some environment variables on Windows, but I could not reflect it.

I analyzed the theme. As a result, I found this workaround:
**config.toml**
```
[params]
  env = "production"
```
It may need to be added to the README.

EricStG:
Hugo can read the Windows environment variables just fine, but you can't set the command & run on a single line (unless you use `&&` in between commands)

This will work for cmdline line. I can add a PowerShell version if there's interest.

Fixes #153

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,8 @@ HUGO_ENV=production hugo
 Note: The above command will not work on Windows. If you are running a Windows OS, use the below command:
 
 ```
-hugo --environment production
+set HUGO_ENV=production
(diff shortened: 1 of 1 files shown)


budparr:
Thank you!

PR #277, opened 2020-05-22
Use | relLangURL for the base url in the site-navigation

jmservera:
This way, when you click on the header it maintains you in the same language you previously selected.

### DIFF
--- a/layouts/partials/site-navigation.html
+++ b/layouts/partials/site-navigation.html
@@ -1,6 +1,6 @@
 <nav class="pv3 ph3 ph4-ns" role="navigation">
   <div class="flex-l justify-between items-center center">
-    <a href="{{ .Site.BaseURL }}" class="f3 fw2 hover-white no-underline white-90 dib">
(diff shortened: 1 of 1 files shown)


layouts/partials/site-navigation.html
```diff
@@ -1,6 +1,6 @@
 <nav class="pv3 ph3 ph4-ns" role="navigation">
   <div class="flex-l justify-between items-center center">
-    <a href="{{ .Site.BaseURL }}" class="f3 fw2 hover-white no-underline white-90 dib">
+    <a href="{{ .Site.BaseURL | relLangURL }}" class="f3 fw2 hover-white no-underline white-90 dib">
```
regisphilibert, line 3: Any reason not to use `.Site.Home.RelPermalink` here? For me it's the safest access to the current site's `.Home`.

jmservera: I tested it and works like a charm, updated to .Site.Home.RelPermalink

PR #281, opened 2020-06-01
Replaced deprecated .URL with .Permalink

lenartj:
This was probably missed in #216 and #275


budparr:
@lenartj DId you test this? Menus do not use `.Permalink` they still use `.URL`

lenartj:
Ah my bad. I though I have, but now I can see I was looking at another version. Sorry!

budparr:
No worries. Thanks for wanting to contribute!

PR #283, opened 2020-06-02
Add head partial

AnimMouse:
Just added head partial to the template.
Some scripts need to be at the head, for example, Google AdSense, etc.
My workaround is to paste the script into the baseof.html
Instead of pasting directly to baseof.html, I recommend to just create a partial at the head to paste the scripts, etc. that needs to be at the head.

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -42,6 +42,7 @@
     {{ if eq (getenv "HUGO_ENV") "production" | or (eq .Site.Params.env "production")  }}
       {{ template "_internal/google_analytics_async.html" . }}
     {{ end }}
+	{{ block "head" . }}{{ partialCached "site-head.html" . }}{{ end }}
   </head>
 
   <body class="ma0 {{ $.Param "body_classes"  | default "avenir bg-near-white"}}{{ with getenv "HUGO_ENV" }} {{ . }}{{ end }}">


budparr:
Hi @adambkaplan I agree with the intent, but I think we should call it something else because there is already stuff in the head. We sometimes use "head_additions." @regisphilibert what do you think?

layouts/_default/baseof.html
```diff
@@ -42,6 +42,7 @@
     {{ if eq (getenv "HUGO_ENV") "production" | or (eq .Site.Params.env "production")  }}
       {{ template "_internal/google_analytics_async.html" . }}
     {{ end }}
+	{{ block "head" . }}{{ partialCached "site-head.html" . }}{{ end }}
```
regisphilibert, line 45: This assumes the code added to the given partials by theme users will be the same for every page, we don't know that and should therefor refrain from using `partialCached`.

Users can create partialCached inside their own `site-head.html` partial by creating subpartials.

AnimMouse: TIL the difference between partial and partialCached.

AnimMouse:
Thanks for adding!

AnimMouse:
PS. @mountainbug95 I wonder why you created another pull request instead of using this pull request.

PR #285, opened 2020-06-03
Add head partial

mountainbug95:
Fixes #283


AnimMouse:
Thanks for merging!

PS. @mountainbug95 I wonder why you created another pull request instead of using my own pull request.

PR #290, opened 2020-06-13
Add style for "hamburger" (3 bar) menu in mobile view

sumpfralle:
* add some IDs to elements of the main navigation
* add styles for mobile view (up to 768px wide) for collapsing the menu
  items into a "hamburger" menu

Closes: #166

# Open issues
## Integration of CSS file into the theme
The path and the inclusion of the `hamburger-menu.css` file are obviously wrong.

I would appreciate hints, how to do it properly within the build workflow of the theme. (tbh: I would prefer, if someone else would do this)

## Social links replaced by hamburger menu button
The change hides the social buttons in the mobile view, since their position (top right corner) is occupied by the hamburger menu button.

But maybe hiding these icons should be optional (i.e. customized in the user's CSS) - thus maybe `.top-social-container { display: none; }` should be removed from the CSS file.

But the new CSS class `top-social-container` (or a different name) should stay in order to allow layout changes for all social links at once (this is impossible right now).

## Bar color
Currently the color of the three bars is hard-coded:

```css
.main-menu-icon:after {
  background: lightblue;
}
```

The above works well for a dark background.

I have no idea, how this could be done dynamically according to the configured color styling of the theme. Assigning the `body_classes` CSS classes to the icon would not work, since the bars are visible via their _background_ color (not the foreground color).

Maybe someone has an idea for a better solution?

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -35,3 +35,5 @@ enableRobotsTXT = true
   background_color_class = "bg-black"
   featured_image = "/images/gohugo-default-sample-hero-image.jpg"
   recent_posts_number = 2
+  # TODO: integrate into build procedure
+  custom_css = ["dist/css/hamburger-menu.css"]
--- a/layouts/partials/site-navigation.html
+++ b/layouts/partials/site-navigation.html
@@ -1,16 +1,18 @@
 <nav class="pv3 ph3 ph4-ns" role="navigation">
   <div class="flex-l justify-between items-center center">
-    <a href="{{ .Site.Home.RelPermalink }}" class="f3 fw2 hover-white no-underline white-90 dib">
+    <a href="{{ .Site.Home.RelPermalink }}" class="site-logo f3 fw2 hover-white no-underline white-90 dib">
       {{ with .Site.Params.site_logo }}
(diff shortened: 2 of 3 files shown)


sumpfralle:
Maybe an example visualization helps ...

# Hamburger menu collapsed

hamburger-collapsed

# Hamburger menu expanded

hamburger-expanded

regisphilibert:
Hi @sumpfralle I'm sorry for that huge gap between your response and mine. We're trying and reach PR 0 and styling PRs were not prioritized. 

We have some concerns about the Hamburger implementation such as this one as we cannot make it accessible without adding some JS layer. Avoiding JS being the main motivation behind a CSS only approach, this kind of beats the purpose.
Also while this proof of concept relies on more css rules, we'd rather use the tachyon utility class library the theme is built with. But this would mean rework of several template files and risk confusing upgrading users.

I think your customization of the theme brings great value and users wanting to benefit from it should use it, but I don't favor it being built in Ananke until we reach a critical release (Ananke 3) which will be able to break from current design and implementation.

Again thanks for your suggestion and sorry for not addressing it sooner 🙏

PR #293, opened 2020-06-15
Add support for page-specific extra scripts

sumpfralle:
The new front matter variable `page_scripts` allows to specify
javascript ressources to be loaded in sync or async mode.
This allows javascript code to be loaded, that is only required for a
small set of pages (e.g. showing a map or a slider animation).


regisphilibert:
Sorry for taking so long, I think it's worth thinking about something that could be added globally or per page. Using site.Params or page Front Matter.

Then we could use `.Param "scripts"`

sumpfralle:
Should I implement this (by updating my branch)? Or do you want to do this on your own?

Which names would you suggest? How about the following?
* global: `site.Params.scripts`
* page-specific: `.Params.script`

Issue #298, opened 2020-06-16
Add translation for taxonomy page


peter-jerry-ye:
Original version: Below you will find pages that utilize the taxonomy term “{{ .Title }}”

PR #299, opened 2020-06-16
Add translation for taxonomy page

peter-jerry-ye:
Modification for #298


budparr:
This looks good, thank you.

PR #303, opened 2020-06-19
Add automatic cover image support

### LINKED ISSUE #302: Set featured image automatically if cover image is present
Currently, Anake seems to require featured images to be set in the front matter. This can create problems if a user want to organize images with an article into page bundles, since you can't access page bundle data from the front matter.

Both the internal twitter card and open graph templates solve this by searching for image files with "cover" or "thumbnail" in their names within a page bundle to automatically generate the opengraph and twitter metadata if it's not present in front matter - it would be great if anake also could grab the featured image from a "cover" image, if present.

Here's the current behavior of twitter and open graph:
https://github.com/gohugoio/hugo/pull/5157

iUnknwn:
This adds support for automatic featured image detection, using the simple rule as the twitter and open graph internal template - if we can find a file named "*cover*".jpg it is automatically selected as the featured image if one was not selected in front matter.

This should make working with bundles easier.

(Fixes #302)


budparr:
> As this is the first time we add a returning partial to the theme
I would also add the partial should be heavily commented, keeping with the intent of this theme to teach.

regisphilibert:
Good call @budparr. Here's a commenting example: 
https://github.com/theNewDynamic/hugo-module-tnd-func/blob/master/functions/IsExternalURL.html

iUnknwn:
As requested, added documentation to the partial and changed the function name.

PR #317, opened 2020-07-18
Fix '+' character being written as &#43;

Bas-Man:
The existing template code results in the + chartering being written in a form that is not valid for the datetime element when used in and <time> tag.

This patch fixes this issue.
references:
https://discourse.gohugo.io/t/date-lastmod-return-43-in-place-of/27033
https://github.com/gohugoio/hugo/issues/7488

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -30,7 +30,7 @@
       </p>
       {{ end }}
       {{/* Hugo uses Go's date formatting is set by example. Here are two formats */}}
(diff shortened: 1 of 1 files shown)


Bas-Man:
An alternative option which I have just been informed of. Would be to replace the Z with a '-'

See:https://discourse.gohugo.io/t/date-lastmod-return-43-in-place-of/27033/9?u=adamspann

regisphilibert:
Thank you for this @Bas-Man

Issue #238, opened 2019-11-29
Params.description does not display on site header


corch:
On the demo site it appears as though the site description set by params.description in config.toml appears under the site title, however on a fresh install of this theme the description does not display. Is there some other config setting which needs to be turned on to make this work (or how else to get a subtitle to display on the header image?),

alienxp03:
Replacing `.Params.description` with `.Site.Params.description` works.

fpiantini:
I solved copying also the `_index.html` file from `exampleSite/content` of the ananke theme, to my site `content` directory (and obviously customizing it). It seems that something is not working with the `description` parameter in the `config.toml` file, but in any case the `_index.html` overrides it... with some duplication, but also with the added value that you can customize the home page of your site.

PR #348, opened 2020-10-02
Update baseof template

enguerran:
I use this template and I realized via Google Lighthouse that the `<meta name="description" content="">` 
tag was missing. Also, I updated the link to the internal templates doc.

- Update comments about internal templates
- Add missing meta description for SEO

Let me know if the PR needs more work.


enguerran:
This covers https://github.com/theNewDynamic/gohugo-theme-ananke/issues/238

enguerran (commented):
I added a commit to add variable scoping within the `head_additions.html` partial.
See https://gohugo.io/templates/partials/#variable-scoping﻿

PR #354, opened 2020-10-14
Fix: Featured image requires absolute path #233

### LINKED ISSUE #233: Featured image requires absolute path
The `featured_image` parameter can be set system wide in `config.toml` or in the front matter of a specific page. This parameter always expects an absolute path from the site root, and does not handle relative urls.

When using it with a page leaf, I have a `posts` section in `content` with folder named `slug-to-page`, containing both an `index.md` and `featured.jpg`:

```
/content/
  posts/
    slug-to-page/
      index.md
      featured.jpg
```

In the index file, I would like to add

```yaml
featured_image: featured.jpg
```

or

```yaml
featured_image: ./featured.jpg
```

The link to the header background image now trims the `'/'` and then makes the path absolute with `absUrl`.

https://github.com/budparr/gohugo-theme-ananke/blob/15a454034900cc0b969db8af490ff77013757624/layouts/partials/page-header.html#L4

Instead, I would like some logic like:

> If it is an absolute path (starting with '/') then leave as is, otherwise precede relative path with permalink from current page.

Something like this:

```go
{{ if not hasPrefix $feature_image "/" }}
  {{ $featured_image := (path.Join( .Permalink $featured_image) }}
{{ end }}
```

pmorch:
This fixes #233 

I have one hesitation: `func/GetFeaturedImage`  that this modifies is called from 3 files:

* layouts/post/summary-with-image.html
* layouts/partials/page-header.html
* layouts/partials/summary-with-image.html

As you can see there is both a `summary-with-image.html` in `post/` and `partials/`. I was not able to find any reference to `post/summary-with-image.html` though, so I was not able to test that code-path. I could only see `partials/summary-with-image.html` being called:

```
me@machine:~/some/path/themes/ananke» ack summary-with-image
layouts/index.html:29:            {{ partial "summary-with-image.html" . }}
```

In my setup, everything worked fine even if `post/summary-with-image.html` is deleted entirely... So some help in how to setup something up that actually uses `post/summary-with-image.html` would be appreciated. But I don't (yet) see why it shouldn't work there also.

### DIFF
--- a/layouts/partials/func/GetFeaturedImage.html
+++ b/layouts/partials/func/GetFeaturedImage.html
@@ -18,13 +18,21 @@
 {{/* Declare a new string variable, $linkToCover */}}
 {{ $linkToCover := "" }}
 
+{{/* Declare a new string variable, $img, if this is an image resource */}}
+{{ $img := "" }}
+
 {{/* Use the value from front matter if present */}}
 {{ if .Params.featured_image }}
-    {{ $linkToCover = .Params.featured_image }}
-
+    {{ $img = (.Resources.ByType "image").GetMatch .Params.featured_image }}
+    {{ if not $img }}
+        {{ $linkToCover = .Params.featured_image }}
+    {{ end }}
 {{/* Find the first image with 'cover' in the name in this page bundle. */}}
 {{ else }}
-    {{ $img := (.Resources.ByType "image").GetMatch "*cover*" }}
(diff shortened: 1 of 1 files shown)


regisphilibert:
Hi @pmorch and thanks for your contribution.

I'm sorry for being so late on this. The featured image logic will be refactored in the upcoming [3.0.0](https://github.com/theNewDynamic/gohugo-theme-ananke/milestone/3) of the theme. 

We'll include your reflexion in there.

PR #355, opened 2020-10-15
enable hosting the website also on other URLs

jan-xyz:
I just submitted a PR for the language picker to have relative URLs because that was a thing I was working on then. I just recently found that other sections are also using the Permalink.

This PR replaces all usage of `.Permalink` with `.RelPermalink`. I setup a test website for this here: https://5f88ad34a2750b00077a188b--janxyz.netlify.app

Without the change references like links to Posts would point to the production version on `jan-steinke.de` and that was annoying to test new things as URLs needed manual intervention. This should fix that problem.

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -37,7 +37,7 @@
         {{/* Now, range through the next four after the initial $n_posts items. Nest the requirements, "after" then "first" on the outside */}}
         {{ range (first 4 (after $n_posts $section))  }}
           <h2 class="f5 fw4 mb4 dib mr3">
-            <a href="{{ .Permalink }}" class="link black dim">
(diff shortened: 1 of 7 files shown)


budparr:
This looks good, @jan-xyz. I'm going to have @regisphilibert give a quick review and he's out until monday. 

Thanks!

jan-xyz:
Any feedback on this? I am still using my fork to be able to use this also on staging environments and would like to get back to the upstream repo :)

regisphilibert (approved):
Sorry for taking so long! This is better of course!

jan-xyz:
No need to apologize 😊 thanks for looking into it

PR #357, opened 2020-10-19
Add go.mod

### LINKED ISSUE #307: Make it a Module
init the theme as a Hugo Module (Go Module)

zhangyuannie:
Closes #307.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/theNewDynamic/gohugo-theme-ananke
+
+go 1.12


zhangyuannie:
Any chance we can land this? Because #307 is not really solved, since `gohugo-theme-ananke` is not a proper go module yet. Using it as a hugo module will yields a `+incompatible` tag in user's `go.mod`. As a theme that specify module as the recommended way to install, I believe we should address this.

Golang documentation: https://github.com/golang/go/wiki/Modules#semantic-import-versioning
Example implementation that is used in Hugo's documentation: https://github.com/spf13/hyde/blob/master/go.mod

regisphilibert:
This is working as is. An incompatible string is not too armful for me. This might cause problem as people will need to make sure they specify (I think) the right Go Version (here you specify 1.12 but 1.14 is the most recent).

Have you ran tests?

zhangyuannie:
The Go version specified in the file is the lowest Go version supported. 1.12 is used here because that seems to be the lowest version supported for Go/Hugo module (https://gohugo.io/hugo-modules/use-modules).

I have tested with `zhangyuannie/gohugo-theme-ananke` previously and I'm currently on go 1.15.

Although it is working as is, this is not the "correct" way to do things. `incompatible` means that this repo is not a go module. It is like a node project without `package.json`.

regisphilibert:
> Although it is working as is, this is not the "correct" way to do things

I never implied otherwise.

PR #359, opened 2020-10-19
Fix confusion in README.md

zhangyuannie:
~~The user's module system may not be initialised. It should be initialised first, otherwise the user will be prompt with:~~
```
Error: module "github.com/theNewDynamic/gohugo-theme-ananke" not found; either add it as a Hugo Module or store it in "/path/to/dir/themes".: module does not exist
```
The above issue has been fixed by ce48112de29f91330ec4da2bfb7ed14d0ee1a19f, the PR has been updated.

The code snippet did not specify that yaml syntax was used instead of the default toml syntax. Changed it to toml syntax to avoid confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Also includes examples of Hugo Features or Functions:
 - `where`
 - Content Views
 - Partials
-- Template layouts (type "post" uses a special list template, single template,  and a content view)
+- Template layouts (type "post" uses a special list template, single template, and a content view)
 - Tags
 - `len`
 - Conditionals
@@ -51,12 +51,17 @@ This theme uses the "Tachyons" CSS library. This will allow you to manipulate th
 
 ### As a Hugo Module (recommended)
 
-Simply add the repo to your theme option:
+1. Initiate the hugo module system:
 
-```yaml
-theme:
-  - github.com/theNewDynamic/gohugo-theme-ananke
-```
+   ```
+   $ hugo mod init github.com/<your_user>/<your_project>
+   ```
+
+2. Add the theme in your `config.toml`:
+
+   ```toml
+   theme = ["github.com/theNewDynamic/gohugo-theme-ananke"]
+   ```
 
 ### As Git Submodule
 
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ This theme uses the "Tachyons" CSS library. This will allow you to manipulate th
    $ hugo mod init github.com/<your_user>/<your_project>
    ```
 
-2. Add the theme in your `config.toml`:
+2. Add the theme's repo to your `config.toml`:
 
    ```toml
    theme = ["github.com/theNewDynamic/gohugo-theme-ananke"]


README.md
```diff
@@ -51,16 +51,17 @@ This theme uses the "Tachyons" CSS library. This will allow you to manipulate th
 
 ### As a Hugo Module (recommended)
 
-If not already, init your project as Hugo Module:
+1. Initiate the hugo module system if you haven't already:
```
regisphilibert, line 54: Why does this sentence make more sense in your view? Is the "hugo module system" something referenced in the Hugo documentation? For me, and it might be biased, your projet is either a Hugo Module or is not. So I would change the existing sentence as `Init your project as a Hugo Module if you have not already:` but fail to see the benefit of your own sentence.

README.md
```diff
@@ -51,16 +51,17 @@ This theme uses the "Tachyons" CSS library. This will allow you to manipulate th
...
 
-`$: hugo mod init {project_repo_url}`
+   ```
+   $ hugo mod init github.com/<your_user>/<your_project>
```
regisphilibert, line 57: Not every one is using GitHub. But I think the user would benefit from a an example using github (so they don't add they remove the `https://`

README.md
```diff
@@ -51,16 +51,17 @@ This theme uses the "Tachyons" CSS library. This will allow you to manipulate th
...
+   ```
 
-Simply add the repo to your theme option:
+2. Add the theme in your `config.toml`:
```
regisphilibert, line 60: Maybe mention we'll be using the Theme's repo "url" as value for the theme?

regisphilibert (commented):
Thanks! I've had a few comments.

README.md
zhangyuannie: Yes, this is a copy paste from the official documentation: https://gohugo.io/hugo-modules/use-modules/#use-a-module-for-a-theme, but both are good in my opinion, this comes to personal choice, you can revert this line if you wish to do so.

README.md
zhangyuannie: Agree. I find `project_repo_url` a bit vague. The presence of a concrete example is arguably clearer, a person can simply replace `github` by `gitlab`, `bitbucket`, or any repo.

README.md
zhangyuannie: The `theme` of Hugo config specify the location as per their documentation, so I think it is clear, especially with the example below, but we could definitively specify this.

README.md
regisphilibert: Can you confirm a module hosted on bitbucket or gitlab uses the same url syntax ? `{domain}/{user}/{repo_name}` (I've never been able to make sure.

README.md
regisphilibert: As in let's keep it consistent with Hugo's documentation.

README.md
zhangyuannie: Yes

Ref: https://blog.golang.org/using-go-modules

PR #376, opened 2020-12-23
ranged through an array to make social-follow.html more efficient

zjrohrbach:
not sure if this is helpful, but I made the `social-follow.html` more succinct by ranging through all of the repetitive code.

This could probably be improved by making sure the `title` and `aria-label` attributes are correctly capitalized for each link.  I used `humanize` to try to achieve this, but it does not split 'Stack Overflow' into two words or correctly capitalize 'LinkedIn.'

### DIFF
--- a/layouts/partials/social-follow.html
+++ b/layouts/partials/social-follow.html
@@ -5,7 +5,7 @@
 
 {{ range $social_media }}
   {{ $this_key := . }}
-  {{ with $dot_passthrough.Param . }}
+  {{ with $dot_passthrough.Param ( . | lower ) }}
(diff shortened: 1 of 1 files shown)


budparr:
Thanks for this. We'll review right after the holiday.

zjrohrbach:
I think this most recent commit fixes most of the capitalization issues.  The only remaining problem is that 'Stack Overflow' still renders as one word instead of two.

regisphilibert:
Thanks @zjrohrbach and sorry for not addressing your PR sooner as it did make a lot of sense. I just merged #422 which refactors the whole social logic of the theme!

PR #385, opened 2021-01-05
fix: do not render a date if it is not set

dijitali:
Fixes the issue where content without a date set render a date of "January 1, 0001".

As shown here:
image

Hopefully this is acceptable, I aimed for this to do the same as was done here:
https://github.com/theNewDynamic/gohugo-theme-ananke/blob/f5cdf1f3a6ce58040dba752b684988e7251c2e37/layouts/post/summary.html#L2-L6

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -29,10 +29,12 @@
           </strong>
       </p>
       {{ end }}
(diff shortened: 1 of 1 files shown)


regisphilibert:
Addresse by #397

PR #386, opened 2021-01-11
Basic example of multilingual

AdamChlan:
@regisphilibert I have a basic example set up for review. Let me know if there is anything else you'd like to see here.


regisphilibert:
Thanks @AdamChlan I just added a blog post and the homepage. Thanks!

AdamChlan:
Okay, you are welcome. I'll plan on reviewing the other open tasks to see if there are any that seem like good candidates to move on.

regisphilibert:
Thanks please make sure they're not in the 3.0.0 milestone as we'll address those then.

PR #387, opened 2021-01-11
Add optional processing of styles with Hugo Pipes

### LINKED ISSUE #379: Consider using Hugo Pipes for the CSS and adding functionality for custom CSS overrides
Hello,

Since this theme is used for the Hugo Quick Start Guide please do consider updating the theme to use current Hugo features like Hugo Pipes for the CSS and not the output of a Webpack workflow. ref: https://github.com/theNewDynamic/gohugo-theme-ananke/blob/master/layouts/_default/baseof.html#L18-L21

Also please consider adding the ability for newbies to easily add Custom CSS .

Currently it is very difficult for new users to override the theme CSS with the present setup, as shown in this [forum topic](https://discourse.gohugo.io/t/complete-beginner-in-website-and-cannot-follow-mathjax-tutorial/30218/6).

regisphilibert:
This bump the min version to .79 but could be lower once we discuss the scripts treatment @budparr 

Fixes #379


regisphilibert:
We end up only bumping to 64.

PR #389, opened 2021-01-15
Change read more to read on because read more is not a descriptive link text for Lighthouse 

Independent-Eye2446:


budparr:
looks good to me. https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/audits/seo/link-text.js

budparr:
Do we need to do this for translations? If so, how?

Independent-Eye2446:
I don't think so because Lighthouse just doesn't like the word ``more``

regisphilibert:
@budparr i leave this one up to you.

budparr:
My reading is that this change will only fix the issue in a technical way. I think that what Lighthouse is looking for is more description links, like `read "title of article"`

PR #390, opened 2021-01-16
Transform the robots meta tag to lowercase (closes #378)

### LINKED ISSUE #378: robots metatag in uppercase
Just wondering, is there a reason why this whole line is in uppercase?

https://github.com/theNewDynamic/gohugo-theme-ananke/blob/master/layouts/_default/baseof.html#L15

DanielRuf:
This PR transforms the robots meta tag to lowercase as browsers parse and display them only in lowercase.


budparr:
If fine with this PR, though I've never seen that meta tags are case sensitive. Would love to see some documentation on that, if you have it.

regisphilibert:
I have to side with the esthetic of consistent case.

regisphilibert:
I have to side with the esthetic of consistent case.

PR #396, opened 2021-02-19
Update site-navigation.html

romainw:
Using `disableKinds = ["home"]` is causing the loss of the `.Site.Home.RelPermalink` variable, which is the href link used in the site-navigation page. Without it, clicking on the site logo no longer links to the homepage. Using `.Site.BaseURL` instead hopefully addresses the issue.

### DIFF
--- a/layouts/partials/site-navigation.html
+++ b/layouts/partials/site-navigation.html
@@ -1,6 +1,6 @@
 <nav class="pv3 ph3 ph4-ns" role="navigation">
   <div class="flex-l justify-between items-center center">
-    <a href="{{ .Site.Home.RelPermalink }}" class="f3 fw2 hover-white no-underline white-90 dib">
+    <a href="{{ .Site.BaseURL }}" class="f3 fw2 hover-white no-underline white-90 dib">
       {{ with .Site.Params.site_logo }}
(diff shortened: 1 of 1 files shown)


itsjoeoui:
`Error: Error building site: render of "404" failed: execute of template failed: template: 404.html:1:24: executing "header" at <partial "page-header.html" .>: error calling partial: execute of template failed: template: partials/page-header.html:23:9: executing "partials/page-header.html" at <partial "site-navigation.html" .>: error calling partial: "/Volumes/Workspace/Dawson/lsc/themes/ananke/layouts/partials/site-navigation.html:3:21": execute of template failed: template: partials/site-navigation.html:3:21: executing "partials/site-navigation.html" at <.Site.Home.RelPermalink>: error calling RelPermalink: runtime error: invalid memory address or nil pointer dereference`

This apparently also fixes the "404" issue.

regisphilibert:
Usinge the baseURL will not point to the right homepage in a multilingual site. `site.Home.RelPermalink` is the right way for a 95% of Hugo projects. I would argue that a website without a homepage is an edge case. For such edge case, one can resolve to overwrite the `site-navigation.html` partial.

PR #403, opened 2021-02-26
added hackernews

liljenstolpe:
I added a hackernews / ycombinator social follow.


liljenstolpe:
Ok, the hackernews icon is not being desaturated to grey, and is WAY too large.  I'm a bit lost here.  Anyone have a pointer as to what I need to fix?

liljenstolpe:
Another issue which I can fix once I unerstand how to make those changes to _social-following.css - the slack hover color should be #4A154B

layouts/partials/svg/hackernews.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="256px" height="256px" viewBox="0 0 256 256" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" preserveAspectRatio="xMidYMid">
```
CJ-Jackson, line 2: You might want to check the width and height and compared them to the other file..

regisphilibert:
Hi @liljenstolpe  I recently merged #422 which refactors the way socials are handled on the site. It makes it very easy for users to add social services on their project even if it's not built in ananke. You can read more about it here: https://github.com/theNewDynamic/gohugo-theme-ananke#social-follow--share

If want you can submit a new PR. All you have to do is 
1. add your service to: [func/socials/GetBuiltInServicesDefaults](https://github.com/theNewDynamic/gohugo-theme-ananke/blob/master/layouts/partials/func/socials/GetBuiltInServicesDefaults.html)
2. Add the icon (name must match key from above) to `assets/ananke/socials/

NOTE: You should revise this PR's icon, as it does not aligned with Ananke's other built in social icons.

PR #404, opened 2021-03-08
fix : production should Allow and point to Sitemap

zar3bski:
My understanding of how **robots.txt** are interpreted by search engines is that they should be explicitly *allowed* to parse the site ([source](https://developers.google.com/search/docs/advanced/robots/robots_txt?hl=en#allow)). Moreover, Google, Bing uses the `sitemap` entry in **robots.txt**. Here, the path is assembled with `{{.Site.BaseURL}}`


zar3bski:
great theme, by the way! If you are open to contribution, I could add this [photo gallery automation](https://zarebski.io/hugo/#the-custom-layout-where-the-magic-occurs)

PR #407, opened 2021-03-29
Changes .GetMatch Pattern to find bundled assets

sgharms:
When I attempted to use `func/GetFeaturedImage.html`, it did not find my resource, `images/foo-cover.webp`. Updating the match syntax to the pattern expressed here, provided expected behavior:

Per:

https://gohugo.io/content-management/page-resources/#pattern-matching

> .Resources.Match "*sunset.jpg" 🚫

and...

> .Resources.Match "*" 🚫

Are not expected to work

> .Resources.Match "**/sunset.jpg" ✅

which follows this change, does.

### DIFF
--- a/layouts/partials/func/GetFeaturedImage.html
+++ b/layouts/partials/func/GetFeaturedImage.html
@@ -24,7 +24,7 @@
 
 {{/* Find the first image with 'cover' in the name in this page bundle. */}}
 {{ else }}
-    {{ $img := (.Resources.ByType "image").GetMatch "*cover*" }}
+    {{ $img := (.Resources.ByType "image").GetMatch "**/*cover*" }}
     {{ with $img }}
         {{ $linkToCover = .Permalink }}
     {{ end }}


regisphilibert:
Thanks @sgharms can you confirm this works even if the image is not at the root of the bundle?

sgharms:
@regisphilibert Upon a revisit, this PR breaks the existing implementation 😢 . Not desired.

It comes down to this: changing the `.GetMatch` to use `**/*cover*` (as in this PR) fixes my nested asset case (which _does_ work at arbitrary depth), **but** breaks the default.

The globbing pattern doesn't seem to be flexible to do both patterns meaning you break my case (nesting) or you break the status quo. Here's a diff to make either work:

```diff
diff --git a/layouts/partials/func/GetFeaturedImage.html b/layouts/partials/func/GetFeaturedImage.html
index a6219c7..b721831 100644
--- a/layouts/partials/func/GetFeaturedImage.html
+++ b/layouts/partials/func/GetFeaturedImage.html
@@ -23,8 +23,11 @@
     {{ $linkToCover = .Params.featured_image }}
 
 {{/* Find the first image with 'cover' in the name in this page bundle. */}}
+{{/* Prefer a top-level *-cover file over a nested/subdir/*-cover */}}
 {{ else }}
-    {{ $img := (.Resources.ByType "image").GetMatch "*cover*" }}
+    {{ $shallowCoverImgPath := (.Resources.ByType "image").GetMatch "*cover*" }}
+    {{ $nestedCoverImgPath  := (.Resources.ByType "image").GetMatch "**/*cover*" }}
+    {{ $img := cond (eq $shallowCoverImgPath nil) $nestedCoverImgPath $shallowCoverImgPath }}
     {{ with $img }}
         {{ $linkToCover = .RelPermalink }}
     {{ end }}
```

I think this gives us non-breaking functionality and supports subdirectory matching at arbitrary depth.

regisphilibert:
Thanks @sgharms let me think about this.

regisphilibert:
Thanks `**cover*` does the trick! I refactored the featured image logic [here](https://github.com/theNewDynamic/gohugo-theme-ananke/commit/1f11fbe012c8def45b5b9dfa9f6ad4f56b831a97)

Thanks again!

PR #408, opened 2021-03-30
Add default author.

diamonster:
If the author is not specified in the front matter, it is taken from the config file.

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -27,6 +27,7 @@ enableRobotsTXT = true
   filename = "sitemap.xml"
 
 [params]
+  author = ""
   favicon = ""
   site_logo = ""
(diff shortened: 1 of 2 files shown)


regisphilibert:
Hi @diamonster and thanks for this. I seem to have merged your PR a bit hastily. Now that I think about it, it might be interesting to use this feature on on section alone, but if a user adds a default_author, it will mean such author will be displayed on every single page of the site. I don't think this is good.

I think people should use [`cascade`](https://gohugo.io/content-management/front-matter#front-matter-cascade) to batch add an author to a section or the whole site.

As a result, even though the feature is now merged and available, I took the liberty not to include in the release note and delete its mention from the README.

Of course, we can discuss this in this PR or a new one.

PR #413, opened 2021-05-24
Adding mathjax support

KunikaValecha:


budparr:
I don't think we want to add a script to the footer for all users of this theme. Might be better to add some documentation about how to do this, instead.

PR #416, opened 2021-06-02
Fix a bug that CSS is not processed correctly when `ananke_process_css = true` and `HUGO_ENV=production`

ngmy:
Fix a bug that CSS is not processed correctly when `ananke_process_css = true` and `HUGO_ENV=production`.


regisphilibert:
Thanks @ngmy. I think this has been addressed by https://github.com/theNewDynamic/gohugo-theme-ananke/commit/6eb54ab5fd8b4932b0123d1fdf94d8026aa67ef6 which will be part of an imminent release

PR #435, opened 2021-09-12
chore: make site description visible as intended

ivanilves:
I have the same issue as @corch here https://github.com/theNewDynamic/gohugo-theme-ananke/issues/238

`.Params.description` gets ignored by the header templates while `.Site.Params.description` works indeed.

### DIFF
--- a/layouts/partials/page-header.html
+++ b/layouts/partials/page-header.html
@@ -8,7 +8,7 @@
       <div class="tc-l pv6 ph3 ph4-ns">
         {{ if not .Params.omit_header_text }}
           <h1 class="f2 f1-l fw2 white-90 mb0 lh-title">{{ .Title | default .Site.Title }}</h1>
-          {{ with .Site.Params.description  }}
+          {{ with .Site.Params.description }}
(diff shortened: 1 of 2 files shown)


regisphilibert (changes requested):
I believe the intent is the following:

Use `.Params.description` if found,
or fallback on `site.Params.description`.

The easiest way to achieve this is to use `.Param "description"`. [See Param](https://gohugo.io/functions/param/#readout)

emacsen:
I have a fresh installation (March 2022) and this is still not working out of the box with params.description, or any version of that as I can tell.

I've tried params.description, param.description site.params.description, site.param.description, and none of them appear to work with displaying the description in the text, though they do appear in the header metadata of the page.

sebandrei:
I'm having an issue with this as well. The site description is set in the config.toml file but does not display in the header. Is this still an unresolved bug?

sebandrei:
Quick update: I checked out the code posted by regisphilibert and replaced .Params.description with .Site.Params.decription in the partials/site-header.html file. This fixed the issue for me.

PR #440, opened 2021-10-30
Base content title on ContentBaseName

jmooring:
With Hugo v0.88.1 and earlier, ContentBaseName and TranslationBaseName
both returned "index" when creating a leaf bundle with hugo new. This is
a known bug, addressed with https://github.com/gohugoio/hugo/pull/9045.

With this PR:

  hugo new post/my-first-post/index.md

will produce:

  title: "My First Post"

instead of:

  title: "Index"


regisphilibert:
Hi @jmooring and thank you for this.

PR #442, opened 2021-11-05
Update it.toml

DavideCalabria:


i18n/it.toml
```diff
@@ -1,11 +1,12 @@
+
 [more]
 other = "Altro"
 
 [allTitle]
 other = "Tutti {{.Title }}"
 
 [recentTitle]
-other = "Recenti {{.Title }}"
+other = "Articoli Recenti "
```
regisphilibert, line 9: I believe this is also in #443? 🤔

regisphilibert (changes requested):

PR #443, opened 2021-11-05
Update it.toml

DavideCalabria:


i18n/it.toml
```diff
@@ -5,7 +5,7 @@ other = "Altro"
 other = "Tutti {{.Title }}"
 
 [recentTitle]
-other = "Recenti {{.Title }}"
+other = "Articoli Recenti"
```
regisphilibert, line 8: This is not necessarily an "article" this is supposed to say "Recent {name_of_section}"

regisphilibert (changes requested):

PR #456, opened 2021-12-12
Add tachyons-style class overrides for layouts

jpursey:
This change adds overrides for text scale, color, alignment, and padding
in all layouts. This can be configured in site or page parameters, and
can target the default for all layouts (mostly useful for scale and
color), or for a specific layout (home, post, list, page, or terms).

The parameters are named in the following way: <type>_<style>:
- default_scale: Defines default scale for content (aka f5)
- default_color: Defines default color for content (aka mid-gray)
- default_padding: Defines default padding for content (aka ph*/pv*)
- default_align: Defines default alignment for content (aka tc/tj)

This can be overridden for "home", "post", "list", "page", or "terms" by
using instead of "default". If a type-specific style is not specified,
then it will fall back on "default", and if default is not specified, it
will fall back on the Ananke theme default.

Of course, theme users may put whatever classes they want in these
settings, but it still is valuable to specify what will be overridden
from the theme.

### DIFF
--- a/layouts/_default/list.html
+++ b/layouts/_default/list.html
@@ -1,6 +1,10 @@
 {{ define "main" }}
   <article class="pa3 pa4-ns nested-copy-line-height nested-img">
-    <section class="cf ph3 ph5-l pv3 pv4-l f4 tc-l center measure-wide lh-copy mid-gray">
+    {{ $cpadding := $.Param "list_padding" | default ($.Param "default_padding") | default "ph3 ph5-l pv3 pv4-l" }}  
+    {{ $cscale := $.Param "list_scale" | default ($.Param "default_scale") | default "f4" }}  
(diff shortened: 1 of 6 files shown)


danielfdickinson:
I like the concept of making these variables and was planning on doing something like this as well. Thanks!

regisphilibert:
Hi @jpursey and thank you for this. 

I understand the benefit but so many technically savvy relying customization (user must know tachyon) might be better addressed through template overwrites. It introduces complexity to the theme code and impacte future maintenance as many files are impacted. This will be useful for developer (who can easily overwrite templates with their own code) but not so much for non-tech users.

I feel like the majority of users would benefit from a more non-technical customization approach (dynamically generated tachyon classes based on user prefs) which is out of scope for this PR and would be better suited for a future major upgrade of the theme.

jpursey:
Sure, that's fair.

PR #457, opened 2021-12-12
Evaluate .Hugo.Environment variable for defining production build

### LINKED ISSUE #453: Support for .Hugo.Environment variable
Just discovered that the [.Hugo.Environment variable](https://gohugo.io/variables/hugo/) is not evaluated as an alternative to HUGO_ENV environment variable and .Site.Params.env theme parameter. Testing for this condition would be an improvement.

### Example
https://github.com/theNewDynamic/gohugo-theme-ananke/blob/v2.8.0/layouts/_default/baseof.html#L12
```html
      <META NAME="ROBOTS" CONTENT="INDEX, FOLLOW">
    {{ else }}
      <META NAME="ROBOTS" CONTENT="NOINDEX, NOFOLLOW">
    {{ end }}
```

tlindsay42:
Closes #453

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,16 +274,23 @@ Now enter [`localhost:1313`](http://localhost:1313/) in the address bar of your 
 
 ## Production
 
-To run in production (e.g. to have Google Analytics show up), run `HUGO_ENV=production` before your build command. For example:
+To run in production (e.g. to have Google Analytics show up), the built-in [Hugo variable: Environment (`.Hugo.Environment`)](https://gohugo.io/variables/hugo/) will default to `production` when running `hugo` to build the site (whereas `hugo server` defaults to `development`), or this can be configured explicitly via the CLI-tag via `hugo --environment 'production'`, or run `HUGO_ENV=production` before your build command. For example:
 
 ```
 HUGO_ENV=production hugo
 ```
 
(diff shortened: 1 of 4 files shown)


tlindsay42:
Everything looks good with overriding with these changes on my end, but still new to all this and wasn't sure of what exactly this `with` call did: [/layouts/_default/baseof.html#42](https://github.com/theNewDynamic/gohugo-theme-ananke/blob/8845854aa9ab5af79066e840d8c884fbbb8a269d/layouts/_default/baseof.html#L42), nor the proper way to update it if necessary to evaluate the other options for specifying a production environment build. If this statement doesn't need to be updated, then this is ready for review. 

?

regisphilibert:
Hey and thank you, this PR does make sense but this is a dangerous breaking change as this would make every project without a `HUGO_ENV` variable set to be considered `production`. Everybody would have to update their non-production deployed branches, PRs, staging to explicitly set a `HUGO_ENV=development`

This is a default of Hugo that I never quite aligned with. Making every project's environment default to `production` is problematic.

I've thought about creating an `IsProduction` returning partial that would preserve the former behaviour while introducing `hugo.Environment` but cannot think of one. Can you?

tlindsay42:
I can't either, and not accepting this makes sense in retrospect due to the risks. Oh well.

Issue #454, opened 2021-12-12
Rendered site sets social image height to '{{ . }}'


tlindsay42:
Noticed the following warnings in my dev tools console for the GitHub, GitLab, and LinkedIn social icons, but not the Twitter icon: `Unexpected value {{ . }} parsing height attribute.`

![image](https://user-images.githubusercontent.com/11078689/145721171-abb21be6-be8b-4e65-99b1-6fb8cb2a4615.png)

danielfdickinson:
I noticed this issue when validating my site as well (actually it was my link checker that complained because it didn't know how to interpret the style; go figure).

This issue is that most of the svgs are assets and not processed as templates and therefore the Go templates in the svg files stay verbatim.

For example, for LinkedIn at https://github.com/theNewDynamic/gohugo-theme-ananke/blob/c7b9901e01ec3720b47fcef3d33bdbd6ce53f000/assets/ananke/socials/linkedin.svg?short_path=3c310ed#L1:
```html
<svg {{ with .size }} height="{{ . }}" {{ end }} style="enable-background:new 0 0 65 65;" version="1.1" viewBox="0 0 65 65"  xml:space="preserve" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
Notice the ``{{ with .size }}…``

Also on the next line:
```html
  <path d="M50.837,48.137V36.425c0-6.275-3.35-9.195-7.816-9.195  c-3.604,0-5.219,1.983-6.119,3.374V27.71h-6.79c0.09,1.917,0,20.427,0,20.427h6.79V36.729c0-0.609,0.044-1.219,0.224-1.655  c0.49-1.22,1.607-2.483,3.482-2.483c2.458,0,3.44,1.873,3.44,4.618v10.929H50.837z M22.959,24.922c2.367,0,3.842-1.57,3.842-3.531  c-0.044-2.003-1.475-3.528-3.797-3.528s-3.841,1.524-3.841,3.528c0,1.961,1.474,3.531,3.753,3.531H22.959z M34,64  C17.432,64,4,50.568,4,34C4,17.431,17.432,4,34,4s30,13.431,30,30C64,50.568,50.568,64,34,64z M26.354,48.137V27.71h-6.789v20.427  H26.354z" style="fill-rule:evenodd;clip-rule:evenodd;fill:{{ .fill }};"/>
```
Notice the ``fill:{{ .fill }};``

[...]

PR #461, opened 2021-12-22
Remove template language from social svg files

funcsec:
I'm not super familiar with Hugo and the templating language, but it appears that the svg files are not parsed and template code that remains is causing browser debug notifications. 

The suspect code is not in all of the svg files, missing from twitter and facebook svg's. 

Error example: 
`Unexpected value {{ . }} parsing height attribute`

Screenshot_2021-12-21_21-21-14
Screenshot_2021-12-21_21-26-22

Great theme!


danielfdickinson:
This should close #454 .

PR #472, opened 2021-12-28
433 fix tags layout

### LINKED ISSUE #433: Layout suggestions for tags and related content
# about tags

Tags are usually no more than 5, so it is better to arrange them horizontally on both mobile phones and computers

# about related content

I've adjusted it down here with custom_css

```css
code {
  padding: 2px 4px;
  border-radius: 4px
}

.cf {
  text-align: justify;
  opacity: 1.0;
  width: 100%;
  height: auto;
}

.nested-links {
  text-align: justify;
  opacity: 1.0;
  width: 100%;
  height: auto;
}

div {
  opacity: 1.0;
}
```

But the width of the relevant content box is too narrow, how to adjust to the same width as the code box?

fullstack-jack:
Fixes #433

### DIFF
--- a/assets/ananke/css/_styles.css
+++ b/assets/ananke/css/_styles.css
@@ -18,7 +18,3 @@
 .nested-links a{
   overflow-wrap: break-word;
 }
-
-.tags li {
-  display: inline;
-}
--- a/assets/ananke/css/_styles.css
+++ b/assets/ananke/css/_styles.css
@@ -18,3 +18,7 @@
 .nested-links a{
   overflow-wrap: break-word;
 }
+
+.tags li {
+  display: inline;
+}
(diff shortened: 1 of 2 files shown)


regisphilibert:
@budparr this is design

regisphilibert (changes requested):
Hi @fullstack-jack and thank you for this. The theme is using a utility class library called tachyon so we don't need custom css for this fix. You can simply apply `di` class on the li. Could you do so and remove the changes you made to `_styles.css`?

```diff
+ <ul class="pa0">
- <ul class="pa0 tags">
  {{ range .Params.tags }}
+   <li class="list di">
-   <li class="list">
     <a href="{{ "/tags/" | relLangURL }}{{ . | urlize }}" class="link f5 grow no-underline br-pill ba ph3 pv2 mb2 dib black sans-serif">

fullstack-jack:
@regisphilibert Thanks for the review, I didn't know about tachyon so it was good to look into. I've made the changes you suggested.

Issue #467, opened 2021-12-24
Use same template for site-header and page-header


regisphilibert:
Currently we're using two different template files for generating page's headers and home and list headers. This creates confusion and asymmetric behavior.

We should use one template and work conditions in there if some page needs a different treatment.

PR #474, opened 2021-12-31
Standardize featured_image logic

### LINKED ISSUE #473: site-header.html featured image
Noticed that the site-header.html partial's featured image logic is different than the home page and the page-header.html partial. This should call the GetFeaturedImage partial so that the behavior is consistent.

tlindsay42:
Updates the site-header partial's featured_image logic to call the GetFeaturedImage partial so that it's consistent with the homepage and the page-header partial.

Closes #473


tlindsay42:
Just noticed #467, which would ultimately supersede this once implemented

tlindsay42:

PR #477, opened 2022-01-03
Update build environment logic

### LINKED ISSUE #453: Support for .Hugo.Environment variable

tlindsay42:
Closes #453
Relates to #457

* Adds GetEnvironmentName and IsProduction partial functions, which improve DRY
* Add prioritization for determining environment name, which favors the existing methods: the `HUGO_ENV` environment variable and site config `env` theme parameter
* Adds `useHugoEnv` theme parameter to use the value of `.Hugo.Environment` for the build environment name

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,18 +296,49 @@ Now enter [`localhost:1313`](http://localhost:1313/) in the address bar of your 
 
 ## Production
 
-To run in production (e.g. to have Google Analytics show up), run `HUGO_ENV=production` before your build command. For example:
+To run in production (e.g. to have Google Analytics show up), you must set the build environment name. The four options for setting the build environment name (ordered by priority) are:
 
-```
-HUGO_ENV=production hugo
-```
+1. The `HUGO_ENV` environment variable.
 
-Note: The above command will not work on Windows. If you are running a Windows OS, use the below command:
+    Examples:
 
-```
-set HUGO_ENV=production
-hugo
-```
+    **macOS/Linux**
+
+    ```bash
+    HUGO_ENV=production hugo
+    ```
+
+    **Windows Command shell**
+
+    ```bat
+    set HUGO_ENV=production
+    hugo
+    ```
+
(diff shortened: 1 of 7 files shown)


tlindsay42:

regisphilibert:
Thanks @tlindsay42 for all you work this past weeks 🙏 . I feel like this one deserves some thinking though. JFY, we're ignoring but might ponder it for some time.

tlindsay42:
Sounds good. My primary driver for this feature was that I chose to implement my site with the env-specific config files (eg: `./config/_default/config.toml` and `./config/production/config.toml`) because I wanted to define a few env-specific behaviors, such as minification. Given this, I was confused at first when parts of the site ran in production mode when I ran `hugo server --environment production`, but others features didn't. 

To test production mode with this type of implementation, customers have to set both the environment variable and the argument, which isn't that big of a deal, but imo it's annoying and not obvious. 

Since the feature is opt-in only, existing customers that aren't expecting this won't be impacted (like we [previously discussed](https://github.com/theNewDynamic/gohugo-theme-ananke/pull/457#issuecomment-1000491491)). 

I'm open to suggestions if changes are desired. 

I hope you decide to merge it, but understand if the decision is otherwise. Thanks!

rgson:
+1. I just started using this theme and was confused by the `--environment` option not being sufficient.

PR #479, opened 2022-01-05
Fix zh* translation

peter-jerry-ye:
The old translations was "What's this" instead of "What's in this"


regisphilibert:
Hey thanks for this,

It's hard for us to evaluate proper Chinese translation, @TMineCola, as the person who initially created the ZH localization, what do you think?

TMineCola:
> Hey thanks for this,
> 
> It's hard for us to evaluate proper Chinese translation, @TMineCola, as the person who initially created the ZH localization, what do you think?

Looks good to me

PR #481, opened 2022-01-08
Fix support for img width

### LINKED ISSUE #480: figure shortcode respects height, but ignores width
Relates to #451 (couldn't re-open, so opening a new one) and https://github.com/gohugoio/hugo/issues/9226

The figure shortcode respects height value, but ignores width. Figured out that this is because of the use of the `nested-img` [Tachyons selector class](https://tachyons.io/docs/table-of-styles/): 

image

This class doesn't appear to be necessary on any of the example site pages (eg: [/about/](https://github.com/theNewDynamic/gohugo-theme-ananke/blob/57e38aad389fb114d82e98591d64b5abeba758c9/exampleSite/content/en/about/_index.md?plain=1#L6)), and the style declarations don't seem desirable as a default. Recommend removing all references so that achieving a specific image width doesn't require custom CSS. 

Sending a quick PR in case you agree.

Example:

```
{{< figure src="/images/avatars/bio-photo.jpg" height="50" width="50" >}}
```

image

### Hugo version

```text
hugo version
hugo v0.91.2-1798BD3F+extended windows/amd64 BuildDate=2021-12-23T15:33:34Z VendorInfo=gohugoio
```

tlindsay42:
Fixes #480


tlindsay42:

regisphilibert:
@budparr I've got a good feeling about this PR, but I'd rather you and your tachyon experience double check.

tlindsay42:
@budparr?

budparr (approved):
This looks fine to me. I'm trying to think of cases where users would not be using the shortcode and if we need to mention anything in documentation.

PR #483, opened 2022-01-08
Ignore Hugo build lock file

### LINKED ISSUE #482: Ignore Hugo build lock file
Reference: [Hugo v0.89](https://github.com/gohugoio/hugo/releases/tag/v0.89.0)
> Hugo now writes an empty file named `.hugo_build.lock` to the root of the project when building (also when doing hugo new mypost.md and other commands that requires a build). We recommend you just leave this file alone. Put it in .gitignore or similar if you don't want the file in your source repository.

I'd like to add an entry for this to .gitignore so that the file doesn't show up in my index when working on pull requests.

tlindsay42:
Closes #482

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ npm-debug.log
 /junit.xml
 partials/structure/stylesheet.html
 
+# Hugo
+.hugo_build.lock


tlindsay42:

PR #512, opened 2022-02-09
parameterise colors and fonts

tstephen:
I have replaced hard-coded fonts and colors with existing params and in one case introduced a `secondary_text_color` to allow things like `RECENT POSTS` to be differentiated from main headings.

Hope you find this acceptable.

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -7,7 +7,7 @@
   {{ $section := .Site.GetPage "section" .Section }}
   <article class="flex-l flex-wrap justify-between mw8 center ph3">
     <header class="mt4 w-100">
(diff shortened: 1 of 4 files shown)


budparr:
Thanks, @tstephen I think this is fine. I've asked @regisphilibert to have a quick look.

regisphilibert:
Hi @tstephen and thank you for this.

The purpose of the `post_content_classes` parameter is to customize the content of a post. By content we mean the body, not the title or the time. This current implementation apply the given classes to all those elements.

In the future we'll think of a better way to customize various fonts of the site, but I feel like this is not the right approach.

Now as for the introduction of `secondary_text_color`, I feel like we should discuss this on a ticket first in order poll the need for a second color and spec some directions for it.

PR #525, opened 2022-03-09
Improve Portuguese translation

LITUATUI:
It doesn't make sense to use "Todo o", we get "Todo o Publicações", "Todo o Notícias" ou "Todo o Posts". 

"Ver mais" (See more) makes more sense, since it can be correctly used for whatever we call our posts in Portuguese. "Ver mais Publicações", "Ver mais Notícias" ou "Ver mais Posts".

### DIFF
--- a/i18n/pt.toml
+++ b/i18n/pt.toml
@@ -2,7 +2,7 @@
 other = "Mais"
 
 [allTitle]
-other = "Todo o {{.Title }}"
+other = "Ver mais {{.Title }}"
 
 [recentTitle]
 other = "{{.Title }} Recentes"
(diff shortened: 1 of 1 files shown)


regisphilibert:
Hi @LITUATUI and thank you for this.

I understand the predicament here for languages where the definite article agrees with the gender of its noun. In this context, `los` (spanish) or `o` (portuguese) will only work with masculine nouns.

But the phrase chosen here is "See all ..." not "See more ..." and the translation key is `allTitle`, so I'm afraid your concern is bigger than just one language.

Can you open a ticket to suggest a wording that would work in more languages, (`See more` is good I think but we'd need confirmation from other translators) so we can work on solving this problem globally?

This way we can update the wording for all languages, along with a translation key which agrees with it.

Thanks again!

PR #532, opened 2022-03-30
Added support for Right-To-Left languages

MeirKriheli:


regisphilibert:
Hi @MeirKriheli this looks great thank you! 

Do you have a repo with a language where this would apply? I'm curious to test it against the default

MeirKriheli:
Not yet, I'm in the process of converting my blog from self written generator to Hugo, attached a couple of screenshots. Also working on adding Hebrew translation.
image
image

How about adding it to `exampleSite`?

regisphilibert:
> How about me adding Hebrew content to exampleSite?

I'd rather not have to many languages in there, but I have my own sandbox where I stress test the theme with some edge cases, if you could share some hebrew md files with me, that would be great! 

Thanks

MeirKriheli:
Here's the WIP of the website I'm working on:
https://github.com/MeirKriheli/meirkrihelicom

regisphilibert:
Hi @MeirKriheli thanks for sharing this repo. I realize the `languageDirection` setting has only been introduced in [Hugo 0.67.1](https://github.com/gohugoio/hugo/releases/tag/v0.67.1). This theme currently as min version support of `64`.

As a result running a previous version, `$.Site.Language.LanguageDirection` will break. We have several options:

- Wait for #541
- Wrap your evaluation of the site's language direction in a partial which will check for the version before using the `.LanguageDirection` method. And use given function throughout (See `/func` directory for code examples)

The latter is great as updating the language direction evaluation logic will be very easy to update in the future.

Let me know if you need any pointer for the function but I guess `{{ if ge hugo.Version "0.67.1" }}` will be plenty.

Thanks!

MeirKriheli:
@regisphilibert added `language-direction` partial. Is that what you had in mind?

regisphilibert:
Thanks a lot, I revisited your logic a tiny bit without notifying you as was in a the middle of publishing a release and felt this had to be part of it.

I made the dir attribute only added if the languageDirection is explicitly set, so no user gets any surprise. https://github.com/theNewDynamic/gohugo-theme-ananke/commit/c49009b823de58ec21e4038eec91329cb538f8d2

Thanks a lot for your work!

PR #533, opened 2022-04-01
Update sv.toml

besynnerlig:
Translated the placeholder [taxonomyPageList] into swedish.


regisphilibert:
Thank you for this @besynnerlig

PR #536, opened 2022-04-10
fix description not showing up in header

cbrunet:


regisphilibert:
Hi @cbrunet and thank you for this.

This does mean that from now on every pages except those using `page-header` (_default/single.html, post/single.html, 404.html) will be using the site's description and users will have no way to have it superseded by the page's own description.

But this brings an interesting point (not the first time #435) because for now, users have no way to add a global and default header description, they need to copy/paste it along pages.

We could simply advise them to rely on the cascade feature of Hugo. @budparr what do you think?

PR #537, opened 2022-04-10
use time.Format to format dates. It allows dates to be localized.

cbrunet:


cbrunet:
Can I just keep the original date format string? I think this should be compatible with older Hugo version, while allowing the user to provide a predefined layout when using a more recent version of Hugo.

regisphilibert:
> Can I just keep the original date format string?

Yes of course! Can you confirm I understand right: That the default would remain "January 2, 2006" as before, and user would be informed to set `:date_full` in the theme's `date_format` params to benefit from new localized dates?

Issue #362, opened 2020-10-31
Make background image responsive


peter-jerry-ye:
There's only one url for each image in css. Is it possible to add media query for responsive images? Does it require a standalone css file?
Thanks.

bennoOoOOooOoOoo:
Need this :)
me too, me too :D

JonathanDoughty:
I would really like to see this too, any more recent thoughts? 

I hacked together support for media queries for the [background/cover](https://jwdoughty.com/post/header-background-images/) image as well as in [summary-with-images](https://jwdoughty.com/post/more-responsive-images/) in an earlier version of ananke. I used an [image shortcode](https://jwdoughty.com/post/adapting-images-to-devices/) as a replacement for hugo's figure too. Now I am struggling to work off a couple years technical debt from not having kept up with ananke/hugo changes in the meantime. That debt has me marooned in hugo 92 and an ananke from around 2.6.7. 

Some built-in ananke support for these would go a long way to making ananke truly 'responsive', show off some of hugo's great [image processing](https://gohugo.io/content-management/image-processing/#image-processing-methods),  and maybe get this hacker out of the business of fumbling his way through go/hugo when I am sure there are more knowledgeable ways of doing these.

nedjo:
A PR is ready for review at https://github.com/theNewDynamic/gohugo-theme-ananke/pull/540.

davidsneighbour:
I will rework the way the featured image is handled. After that this is either magically working or will receive a solution. For a quick manual fix have a look at the media break points Tachyons hands you and add a prepared URL for each breakpoint. My guess is that the final solution will have something like this:

```css
  @media (--breakpoint-not-small) {
    .background-image-selector {
      url(...)
    }
  }

  @media (--breakpoint-medium) {
    .background-image-selector {
      url(...)
    }
  }

  @media (--breakpoint-large) {
    .background-image-selector {
      url(...)
    }
  }
    .background-image-selector {
      url(...) <-- everything smaller than "not-small"
    }
```

PR #540, opened 2022-04-11
Make images responsive (#362)

nedjo:
This PR makes images responsive in three contexts:

* `featured_image` as rendered in summary pages that include images (`summary-with-image`).
* `featured_image` as used in a background in the header.
* Images added via the `figure` shortcode.

It's in response to issue #362, which is focused on the header background image. However, making background images in the header responsive required changes to the `GetFeaturedImage` partial, also used in `summary-with-image`; and the changes needed for `summary-with-image` were readily applicable to the `figure` shortcode.

Two different approaches are taken to responsive images:

1. A set of image URIs via a `srcset` attribute on `` elements. This is the preferred approach and used where feasible, with summary images and `figure`.
1. Media queries that use breakpoints. This fallback approach is used in the header, where we don't have an `` element to work with.

## Activating responsive images

After applying this change, the minimum change needed to get responsive images working is to specify a set of `breakpoints` in the site config file--see details below. Note that, because per [Hugo image processing documentation](https://gohugo.io/content-management/image-processing/) "To process an image, you must access the image as either a page resource or a global resource," only images that are resources will be responsive.

## New `GetImageSrcset` partial

This func-type partial is used internally by both the `figure` shortcode and the `summary-with-image` partial. Sample usage:

```
{{ $srcset := partial "func/GetImageSrcset" (dict "original" $featured_image "target_width" $target_width) }}
```

`featured_image` here is the featured image and `target_width` the expected maximum width available to present the image, expressed as a ratio of the total width at a given breakpoint.

## New override to default Hugo `figure` shortcode

A new `figure` shortcode adds a breakpoint-based `srcset` attribute to images embedded in posts.

## Optional breakpoints

This optional parameter can be provided in the site's config file to supply a set of breakpoints. Only if it is present will images be responsive.

  ```toml
# Use defaults from tailwindcss, https://tailwindcss.com/docs/responsive-design.
breakpoints = [640,768,1024,1280,1536]
```

## New `featured_image_data` parameter in post front matter

This optional parameter can be used to specify an `anchor`, used as a setting when cropping the featured image for use in the header. See valid values at https://gohugo.io/content-management/image-processing/#anchor.

Example:
```yml
featured_image_data:
  anchor: "Center"
```

## Explicit `alt` attribute

As well as `anchor`, the `featured_image_data` parameter can be used to provide an `alt` value to be used as alt text for the image.

Of course, alt text is tangential to the focus of this PR, but it was not possible to do the work in this issue without touching existing code that generates the alt text.

```yml
featured_image_data:
  anchor: "Center"
  alt: "Illustration from Victor Hugo et son temps (1881). Drawing of the Hunchback of Notre Dame, showing the recently restored galerie des chimères."
```
The PR also provides a missing internationalized string for the fallback if an `alt` value is not provided.


regisphilibert:
Hi @nedjo and thank you for this.

I think the logic for images using srcset is sound and limiting it to the direct input of the user (by setting breakpoint) prevents breaking changes and unexpected behaviour.

I don't think we're ready to use the solution applied for background images though. I think this should be wrapped in its own function and asset template so does not crowds the templates. In any case I'd like to give it more time.

Now this breaks with the example site, could you address it before we move into a more in depth review? Here's the error:

```
Error: Error building site: failed to render pages: render of "page" failed: execute of template failed: template: page/single.html:1:24: executing "header" at <partial "page-header.html" .>: error calling partial: execute of template failed: template: partials/page-header.html:4:6: executing "partials/page-header.html" at <partial "header-image-style.html" .>: error calling partial: "/opt/build/repo/layouts/partials/header-image-style.html:5:51": execute of template failed: template: partials/header-image-style.html:5:51: executing "partials/header-image-style.html" at <$featured_image.RelPermalink>: can't evaluate field RelPermalink in type string
```

Thanks a ton!

nedjo:
@regisphilibert 

Thx for the review!

> I don't think we're ready to use the solution applied for background images though. I think this should be wrapped in its own function and asset file (to use with . ExecuteAsTemplate) so it does not crowd the templates. In any case I'd like to give it more time.

k, I'll try to see if I understand what's involved. Any hints appreciated.

> Now this breaks with the example site

Yeah, on the site I developed this on I'm only using page resource images. I'll need to rework so it also supports static image files, and ensure the example site includes both image types.

layouts/partials/func/GetFeaturedImage.html
```diff
@@ -1,39 +1,38 @@
...
+{{/* Declare a new variable, $coverImage */}}
+{{ $coverImage := false }}
 {{ $matches := "feature,cover" }}
 {{/* Use the value from front matter if present */}}
 {{ with .Params.featured_image }}
-	{{ $linkToCover = . }}
+	{{ $coverImage = . }}
   {{/* If we find a Page Resource matching the exact value, we use it instead. */}}
   {{ with $.Resources.GetMatch . }}
-    {{ $linkToCover = .RelPermalink }}
```
regisphilibert, line 26: This is what breaks, this function is expected to return a string, but you store a resource (same below). This will also involves some rewriting of templates which are invoking .RelPermalink on the function's returned value.

nedjo:
> I'll need to rework so it also supports static image files

By which I guess I mean, since we [can't do processing and hence can't use a `srcset`](https://gohugo.io/content-management/image-processing/) without a resource, at least it doesn't _break_ with static image files.

layouts/partials/func/GetFeaturedImage.html
nedjo: Yeah, one way or another we need to distinguish between a featured image that's resource and one that's a static file (string). One way would be splitting this into two different functions, maybe `GetFeaturedImageResource` and `GetFeaturedImagePermalink` or similar and calling them successively. Alternately I guess we test the return value.

nedjo: @regisphilibert 

I've made a number of changes/fixes.

> I don't think we're ready to use the solution applied for background images though. I think this should be wrapped in its own function and asset file (to use with . `ExecuteAsTemplate`) so it does not crowd the templates.

I had a go at this in https://github.com/theNewDynamic/gohugo-theme-ananke/pull/540/commits/fd9ed3b10a8daf0a8bbd100f0ac64bf715b76c63. Is that kinda what you had in mind? Further suggestions?

> one way or another we need to distinguish between a featured image that's resource and one that's a static file (string).

This is still todo.

nedjo: >> one way or another we need to distinguish between a featured image that's resource and one that's a static file (string).

> This is still todo.

Fixed in https://github.com/theNewDynamic/gohugo-theme-ananke/pull/540/commits/cfca913459e9a80966f515d3d5f165004ca35b87. The approach is:

* Output a featured image if either (a) there was a featured image resource or (b) `.Params.featured_image` is not empty.
* Use `GetFeaturedImage` to look for a featured image resource. If none found, use `.Params.featured_image` as the image link.

Sample usage (from `summary-with-image.html`):

```
{{- $featured_image := partial "func/GetFeaturedImage.html" . -}}
[...]
      {{- if or $featured_image .Params.featured_image -}}
        {{- $default_image_url := "" -}}
        {{- with $featured_image -}}
          {{- $default_image_url = .RelPermalink -}}
        {{- else -}}
          {{- $default_image_url = .Params.featured_image -}}
        {{- end -}}
```

regisphilibert:
Thanks! This is a lot fo changes and a big review for me, so I have to push it to the next release! hope to get some time to take a good look at it soon! Thanks

nedjo:
> Thanks! This is a lot fo changes and a big review for me, so I have to push it to the next release! hope to get some time to take a good look at it soon! Thanks

Thx for the update, sounds good.

Meantime I've resolved a few merge conflicts from the changes in https://github.com/theNewDynamic/gohugo-theme-ananke/commit/19242fd2f89b9ac4d469ebbe265808fae17a2388.

nedjo:
There's a new round of conflicts from the changes in https://github.com/theNewDynamic/gohugo-theme-ananke/commit/470ea40982f5036554819253c3ac6ed4a34193f4. I'll have a go at resolving them when I find time.

PR #567, opened 2022-08-18
Add rel=me attribute to social links: fixes mastodon account metadata verification

### LINKED ISSUE #417: add rel="me" attribute option to socials

Altonss:
Adds rel=me attribute to social links
This fixes #417 about mastodon account metadata verification

### DIFF
--- a/layouts/partials/social-follow.html
+++ b/layouts/partials/social-follow.html
@@ -1,7 +1,7 @@
 {{ $socials := where (partialCached "func/socials/Get" "socials/Get") "follow" "!=" false }}
 <div class="ananke-socials">
   {{ range $socials }}
-    <a href="{{ .url }}" target="_blank" class="{{ .name }} ananke-social-link link-transition stackoverflow link dib z-999 pt3 pt0-l {{ cond (eq $.Site.Language.LanguageDirection "rtl") "ml1" "mr1" }}" title="{{ .label }} link" rel="noopener" aria-label="follow on {{ .label }}——Opens in a new window">
+    <a href="{{ .url }}" target="_blank" class="{{ .name }} ananke-social-link link-transition stackoverflow link dib z-999 pt3 pt0-l {{ cond (eq $.Site.Language.LanguageDirection "rtl") "ml1" "mr1" }}" title="{{ .label }} link" rel="me noopener" aria-label="follow on {{ .label }}——Opens in a new window">
       {{ with .icon }}
         <span class="icon">{{ . }}</span>
       {{ else }}


regisphilibert:
Thanks for this @Altonss, [TILT for me](https://www.smartinsights.com/search-engine-optimisation-seo/index-inclusion/google-rel-me/). Now I'm not sure this is the right use case for a company using the theme though, as the `rel=me` attribute has been written for a person. I'm confident Ananke is uses for persons mostly, but there has to be some companies using it...

What do you think?

Altonss:
> What do you think?

I'm not an expert at all, but reading the conclusion of the blogpost you just linked:
```It’s worth noting that this is a work around, and the real [purpose of the 'rel=me'](http://microformats.org/wiki/rel-me) relationship attribute is to join up a person’s identity and not a ‘thing’s’ identity (in this case a company / brand). It will be interesting to see if in future, a new brand identity attribute is launched for this exact use, in response to the large adoption of company social media profiles over the last few years.```
I think this is a legit workaround. But the article seems to be from 2013 so maybe we could look for up to date infos ^^

regisphilibert:
Okay so I did find more recent [documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel). Curiously enough [this other one from mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel) is saying the `me` value is reserved for `link` tags (not `a` tags)... So a bit of  contradiction from mozilla...

In any case, I've discussed this lenghily with @budparr and here is our conclusion:

Even though Google tried to popularise the use of `rel=me` for SEO purposes or simply identifying your social circle, it clearly did not take off, considering the age of the related posts we were able to find on the matter and a more human factor: We would know it by now, if this was important for SEO... 

Now it seems to be used for validating a website's ownership ([RelMeAuth](https://microformats.org/wiki/RelMeAuth)). But for that relevant use case I don't think the theme should enable `rel=me` by default. I would advise any user in need of this to overwrite the `baseof.html` and and insert a `<link rel="me" src="mastondon...">`. I think it's worth the extra step.

nemobis:
Using rel=me is important, inter alia, for verification on Mastodon. That's becoming very mainstream now [for obvious reasons](https://grahamcluley.com/mastodon-what-you-need-to-know-for-your-security-and-privacy/).

The original idea https://microformats.org/wiki/rel-me was that this would be a bidirectional relationship, which implies the rel=me should ideally be only on the canonical about page or mainpage (example: https://github.com/publiccodenet/jekyll-theme/pull/101/files ), but that's not too strict.

PR #568, opened 2022-08-21
Fix font hardcoded!

### LINKED ISSUE #561: Font classes not applied consistently
Hello,

I tried to switch to a sans-serif font, so I added
```
params:
  body_classes: "sans-serif"
  post_content_classes: "sans-serif"
```
That works in most places but not everywhere. 

On the article overview, the caption is still serif:
image

Same for the caption in post details:
image

Altonss:
Fixes #561


regisphilibert:
Hi @Altonss and thank you for this. Even though it's a good concern as users should have full control of the fonts used on the site, this change would alter the styling of existing users (who did not set any `body_classes` or `post_content_classes`.

So we will not merge it. But I'll create a ticket to better address font styling

Issue #546, opened 2022-04-25
Investigate switching from `_internal/google_analytics_async.html` to `_internal/google_analytics.html`


regisphilibert:
According to [Hugo Doc](https://gohugo.io/templates/internal#use-the-google-analytics-template), we should probably use `_internal/google_analytics` which supports G4.

I have yet to understand the consequences of such a switch.

budparr:
very bad idea to not use async

Aweponken:
> very bad idea to not use async

Could you elaborate a bit more on this comment @budparr?

budparr:
Perhaps I'm missing something about the loading of G4, but unless it's loaded async automatically, I would never load a blocking script (except for certain, typically rare instances).

PR #572, opened 2022-08-30
Add support for Google Analytics 4

steven-solomon:
This PR makes the following changes
- Replaces async GA template with `_internal/google_analytics.html`

## Context 
The [Hugo documentation](https://github.com/gohugoio/hugo/blob/ffbdcc75ab7d1d96219e84a20e53de9358fe5adf/docs/content/en/templates/internal.md?plain=1#L59) recommends using `_internal/google_analytics.html` for GA4.

closes https://github.com/theNewDynamic/gohugo-theme-ananke/issues/546

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -37,7 +37,7 @@
     {{- template "_internal/twitter_cards.html" . -}}
 
     {{ if eq (getenv "HUGO_ENV") "production" | or (eq .Site.Params.env "production")  }}
-      {{ template "_internal/google_analytics_async.html" . }}
+      {{ template "_internal/google_analytics.html" . }}
     {{ end }}
 	{{ block "head" . }}{{ partial "head-additions.html" . }}{{ end }}
   </head>


layouts/_default/baseof.html
```diff
@@ -37,7 +37,7 @@
     {{- template "_internal/twitter_cards.html" . -}}
 
     {{ if eq (getenv "HUGO_ENV") "production" | or (eq .Site.Params.env "production")  }}
-      {{ template "_internal/google_analytics_async.html" . }}
+      {{ template "_internal/google_analytics.html" . }}
```
steven-solomon, line 40: I see that the root issue expresses concerns about this change. Should I place this behind a flag?

jhenstridge: You can see the source for the two templates here:

https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/google_analytics.html
https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/google_analytics_async.html

The concerns seemed to be about switching to non-async loaded JavaScript, but it is pretty clear the both templates are async now. It's not obvious the async template was ever any more "async": just a little less obfuscated.

jhenstridge (approved):
(I'm not an Ananke developer, so my approval isn't enough to get this merged)

Here is the v3 analytics code in the `google_analytics.html` template:

```javascript
	(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
	(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
	m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
	})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
```

The question was whether this is async or not. To help determine that, here's my attempt at un-minifying the code:

```javascript
{
    window['GoogleAnalyticsObject']='ga';
    window.ga = window.ga || function() {
        window.ga.q = window.ga.q || [];
        window.ga.q.push(arguments);
    }
    window.ga.l = 1*new Date();
    let a = document.createElement('script');
    let m = document.getElementsByTagName('script')[0];
    a.async = 1;
    a.src = 'https://www.google-analytics.com/analytics.js';
    m.parentNode.insertBefore(a,m);
}
```

So it is loading the same JS blob from Google as the async template does, and loads it asynchronously. It also sets up the `window.ga` object in the same way.

Looking at Hugo's git history, it looks like it has been like this since at least May 2018 (when the internal templates were reorganised), and likely a lot longer.

steven-solomon:
@regisphilibert, would you be willing to share your thoughts here too?

regisphilibert:
I will next week.

regisphilibert:
I think this is the right decision. I'll include a comment in the release that users not comfortable with this should overwrite the `baseof.html` to use the alternative internal google analytics.

PR #581, opened 2022-10-20
Fix file permissions

compwiztobe:
Someone's on Windows...


regisphilibert:
Hi @compwiztobe could let us know more about the motivation behind this PR. Your comment leaves a lot to interpretation.

compwiztobe:
@regisphilibert pretty much just a style choice.  Files coming from NTFS or FAT usually have the execute bit set (even when they originally came from linux), which if you expect it can be handled with clean up scripts or git config.  But I noticed that already this repo has different execute bit values for different values which seem otherwise similar (probably someone on Windows committed these files from an NTFS file system without the relevant git config).  So this change makes them consistent again (removing execute bit because it's not needed) to reverse that corruption and so it's more straightforward to detect/cleanup when your local repo/OS is corrupting it (all or nothing).

Related: https://git-scm.com/docs/git-config#Documentation/git-config.txt-corefileMode

regisphilibert:
Thanks for clarifying and for the work!

PR #587, opened 2022-11-11
Feature: Support rel="me" social links

adambkaplan:
- New social option `verify` which allows a social follow link to be used for RelMeAuth verification.
- Update social-follow.html partial to use `rel="me"` if the `.verify` option is set to `true`.
- Document the `.verify` value in partial functions.
- Update the example `config.toml` to use the verify option.
- Document `verify` option in README.md

Fixes #417

PR #591, opened 2022-11-21
layouts: add support for setting additional relations on social links

### LINKED ISSUE #417: add rel="me" attribute option to socials

jhenstridge:
This is an attempt to allow setting rel="me" on social links in the page header/footer. I hadn't noticed #587 when writing this, so consider it an alternative proposal to fix the problem.

This PR adds a new `rel` attribute to services listed under `ananke_socials`. If set, it specifies additional relations that will be added to the corresponding social link. This should cover case of claiming ownership of a Mastodon account, while also handling any other relations the user might want to add:

```yaml
ananke_socials:
  - name: mastodon
    url: ....
    rel: me
```

While I'm just using a simple string property, it's possible to add multiple new relations by treating it as a space separated list (e.g. `rel: one two three`). When used, it adds to the existing `noopener` relation rather than replacing it.

Fixes #417

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,18 @@ params:
 On top of easily customizing the built-in services' label and color, user can overwrite their icon by adding an svg file at `/assets/ananke/socials` with a filename matching the service's name.
 For example, in order to use your own GitHub icon, simply add an svg file at `/assets/ananke/socials/github.svg`
 
+#### Additional relations
+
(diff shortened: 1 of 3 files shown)


regisphilibert:
Terribly sorry I did not see his earlier... I rushed into fixing #417 myself without checking contributions... You'll find my solution is very similar to yours though.

Thanks a lot for this is much appreciated, and sorry again!

PR #596, opened 2022-12-02
Add a `rel` option to social registers

### LINKED ISSUE #417: add rel="me" attribute option to socials

regisphilibert:
Fixes #417


Altonss:
Thanks a lot!

adambkaplan:
@regisphilibert I don't think this produces valid HTML - see https://stackoverflow.com/questions/39890586/multiple-rel-attributes.

You can only have one `rel` attribute, but you can have multiple values within.

regisphilibert:
What a day I'm having... thanks for pointing this out @adambkaplan. Forgot about the noopener...

PR #597, opened 2022-12-06
assets: update mastodon logo in social links

jhenstridge:
The SVG currently used for Mastodon social links is not the Mastodon logo. From what I can tell, the one in use is an unofficial one from https://openclipart.org/detail/301022/new-logo-alternative-for-mastodon-social. I'm not sure what it's provenance is beyond that. Further more:

1. It isn't monochrome like the other links, so always has 3 white dots in the speech balloon.
2. It doesn't have the same padding as the others, so appears larger when placed beside other social links.

This PR replaces the SVG with one based on the "Mark only" SVG from https://joinmastodon.org/branding. I've adjusted the size/padding to match the others.

I've also adjusted the default hover colour for the link to use one from their branding guide.

For a visual comparison, here's what it looks like without the changes:
image
... and after they're applied:
image


regisphilibert:
Thank you for this!

PR #601, opened 2023-01-06
example site: set languageCode for french language

muhlemmer:
When using the config from the Example site, the document language is not set correctly for french.:

```
<html lang="en-us">
```

The site's language code needs to be set explicitly to fix this, it seems.


regisphilibert:
Fixed with #615

PR #603, opened 2023-01-21
[Fix]: Navigation menu items URL

### LINKED ISSUE #604: Main menu links are incorrect under certain circumstances
Hi, I have a situation where links in the main navigation coming from config are not correct:

- multilingual setup with defaultContentLanguageInSubdir
- baseURL has a subdirectory
- Most menu items come from frontmatter
- A few menu items come from menus.main in config
- hugo v0.109.0

For frontmatter menu items, your site_navigation template generates the correct links, because .URL is already the .RelPermalink. The URLs from manual config items are missing the language prefix and need a relLangURL or absLangURL pipe which prepends it.

tamaracha:
This fixes #604. Please see gohugoio/hugo#4995 for further explanations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,5 +30,22 @@ npm-debug.log
 /junit.xml
 partials/structure/stylesheet.html
 
-# Hugo
+# Created by https://www.toptal.com/developers/gitignore/api/hugo
+# Edit at https://www.toptal.com/developers/gitignore?templates=hugo
+
+### Hugo ###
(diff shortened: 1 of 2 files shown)


regisphilibert:
Hi and thank you for this. As this theme's purpose is also educational, we feel like this would complexify the code for an edge case.

But if you have a repo somewhere with the setup involved, I'd love to take a look and maybe report the problem upstream.

PR #619, opened 2023-02-04
Decrease the top margin a bit

melroy89:
Decrease the top margin of the side panel on large displays from mt6 to mt3.

Especially having table of contents enabled (`toc: true`) the content of the side panel is not easily identified as such. Actually the TOC is almost off the screen.. :\

So less ideal. In my case people should detect the table of contents easier.

For reference, I'm using a 1920x1080 panel.

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -60,7 +60,7 @@
       </div>
     </div>
 
-    <aside class="w-30-l mt6-l">
+    <aside class="w-30-l mt3-l">
       {{- partial "menu-contextual.html" . -}}
     </aside>
 


melroy89:
Or maybe even better, remove this `mt6-l` all together.

regisphilibert:
@budparr this is some design consideration.

ramiro:
Wouldn't it be better to make the whole aside right side lateral block conditional on the toc frontmatter option actually being enabled?

If, in the demo site, you compare a blog post (e.g. [Chapter VI: Esmeralda](https://gohugo-ananke-theme-demo.netlify.app/post/chapter-6/)) with the [About page](https://gohugo-ananke-theme-demo.netlify.app/about/) on a medium sized computer display (no need for it to be any of the very high DPI panels available) you can see the former's body is placed too to the left with too much white space on the right and in comparison the latter looks more pleasantly centered on the screen.

melroy89:
> Wouldn't it be better to make the whole aside right side lateral block conditional on the `toc` front matter option actually being enabled?
> 
> If, in the demo site, you compare a blog post (e.g. [Chapter VI: Esmeralda](https://gohugo-ananke-theme-demo.netlify.app/post/chapter-6/)) with the [About page](https://gohugo-ananke-theme-demo.netlify.app/about/) on a medium sized computer display (no need for it to be any of the very high DPI panels available now) you can see the former's body is placed too to the left with too much white space on the right and in comparison the latter looks more pleasantly centered on the screen.

Both yes. this PR is mainly about the top, but sure also the centering is an issue, when you don't enabled  the toc.

Issue #574, opened 2022-09-07
Support Hugo's built in build environment feature


jhenstridge:
Hugo has built-in support for multiple build environments that is exposed to templates via the [`hugo.Environment` and `hugo.IsProduction` variables](https://gohugo.io/functions/hugo/). This is controlled via the `--environment` command line argument with the `HUGO_ENV` environment variable as a fallback.

Ananke has some support for build environments, but it does it by checking the `HUGO_ENV` environment variable directly:

https://github.com/theNewDynamic/gohugo-theme-ananke/blob/5a8b531a7ce2f22eaa452ebab8325040982c9275/layouts/_default/baseof.html#L12

This means Ananke's view of the build environment will match Hugo's if it was set by the environment variable but not if it was set via the command line.

It looks like this could be simplified to just using `hugo.IsProduction`. This would result in a change in behaviour when no environment is set at all. [According to the docs](https://gohugo.io/getting-started/configuration/#configuration-directory), here are the results:

> Default environments are **development** with `hugo server` and **production** with `hugo`.

regisphilibert:
Exactly, if we do this, then Ananke would treat any website with the default `hugo` command line as a production site. Any deploy preview etc... would be considered production.

It's an easy code change, but a difficult effort in communication to make sure everyone gets the memo.

Juanpam:
@regisphilibert I found this today! I'd say it would be worth it to document it at least as this theme is used in the getting started tutorial from Hugo. It can be confusing since it kinds of conflicts with what's specified in the docs. 

BTW, I got the memo lol

PR #621, opened 2023-02-05
Show how-to use the build-in hugo -e flag

melroy89:
You don't need to set `HUGO_ENV`, you can use the built-in `-e` or `--environment` flag on the `hugo` command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,6 +358,12 @@ set HUGO_ENV=production
 hugo
 ```
 
+Alternatively, you can also run production build using:
+
+```
+hugo -e production
+```
+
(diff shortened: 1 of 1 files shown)


melroy89:
First bug needs to be resolved: https://github.com/theNewDynamic/gohugo-theme-ananke/issues/625

regisphilibert:
You do have to use `HUGO_ENV` for this theme to build the "production" build. See https://github.com/theNewDynamic/gohugo-theme-ananke/issues/574#issuecomment-1297434544 for the motivation behind this.

It's documented.

PR #626, opened 2023-02-21
localization: Create Czech (cs) localization

petrmvala:


petrmvala:
Does it? It doesn't seem to me like an issue for this repository. Why not report it in gohugo repo and use HUGO_ENV until that is resolved?

PR #627, opened 2023-03-01
Support for dynamically localizing menus with i18n

peterjochum:
This PR adds support for an i18n identifier for items in the main navigation. 

This is described in the [Multilingual menus](https://gohugo.io/content-management/multilingual/#multilingual-themes-support) section of the Hugo documentation.


peterjochum:
There remains a problem with the title attribute of the navigation links - it gets the untranslated name of the navigation link.

```
title="{{ i18n "pageTitle" . }}"
```
What is the title attribute for? Aside from adding the information its a "page" it does not contain additional information for users.

image

PR #634, opened 2023-04-27
Include authors meta tag

j0hax:
This patch includes a `<meta name="author" ...` tag. If there are multiple authors, their names will be comma-seperated.

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -16,6 +16,15 @@
     {{ else }}
       <meta name="robots" content="noindex, nofollow">
     {{ end }}
+    {{ with .Params.author | default .Site.Params.author }}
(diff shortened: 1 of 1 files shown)


j0hax:
Hi! According to W3C, `author` is a [standard name](https://html.spec.whatwg.org/dev/semantics.html#standard-metadata-names) for the `<meta>` tag:
> The value must be a free-form string giving the name of one of the page's authors.

More examples from around the web:
- https://webmasters.stackexchange.com/questions/114921/how-should-i-correctly-use-the-meta-name-author-tag
- https://www.ionos.com/digitalguide/websites/web-development/the-most-important-meta-tags-and-their-functions/

I believe that providing machine-readable metadata about the author can improve user experience in e.g. reader views or social share previews.

PR #649, opened 2023-09-15
refactor: use the new twitter (X) logo

Nr18:
Since twitter updated it's logo it makes sense to update it in the theme.


regisphilibert:
The theme is using disk icon with the logo inside... Could you provide something equivalent in size and shape? Thanks.

Nr18:
Graphics is far from my expertise... You want it smaller or in a disk? The GitLab logo isn't in a disk either I noticed...

regisphilibert:
> The GitLab logo isn't in a disk either I noticed...

True, it's not a perfect system, but it still integrates nicely with the other icons. I guess my reluctance to approve this outright is triggered by the size and color of your proposition. 

Capture d’écran, le 2023-12-15 à 13 12 41

Twitter/X is probably on 80% of Ananke sites, so I want to tread lightly.

Thanks for your understanding.

PR #651, opened 2023-09-15
feat: show featured image in summary (when available)

Nr18:
Display the feature image as a preview when displaying a list of pages.

Example can be found here: https://conijn.io/blog/

I do not have a deep understanding of Hugo templating, so might be nicer to create an option to enable it per content type?


regisphilibert:
@budparr this is a design take

PR #655, opened 2023-10-01
Use of .RelPermalink brokes the multilingual host functionality

ricard2k:
https://gohugo.io/content-management/multilingual/#configure-multilingual-multihost

this functionality needs to create the language links with .Permalink


ricard2k:
You can see the patch working in:
https://little-railways.com
https://little-railways.es

regisphilibert (changes requested):
Hi @ricard2k and thank you for this. 

I understand the motivation but I don't think we should enforce `.Permalink` for everyone for such an edge-case.

I'm not sure there is a way to know if the current project is multi host, but if there is, we should test that before using `.Permalink` here rather than `.RelPermalink`.

If this test is not possible then we should add a global boolean setting to the theme `localizationMultiHost`(for example) and test it instead.

Of course the latter would require an addition to the read me.

Thanks a ton!

PR #660, opened 2023-10-24
Access Disqus shortname from canonical location

jmooring:
Prepare for deprecation described [here](https://github.com/gohugoio/hugo/commit/2eca1b3cc1cd03fee40c0abcdc61dab1016c0475).

This is a backwards compatible change.

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -51,7 +51,7 @@
       {{- .Content -}}
       {{- partial "tags.html" . -}}
       <div class="mt6 instapaper_ignoref">
-      {{ if .Site.DisqusShortname }}
+      {{ if .Site.Config.Services.Disqus.Shortname }}
         {{ template "_internal/disqus.html" . }}
       {{ end }}
       {{ if .Site.Params.commentoEnable }}
--- a/README.md
+++ b/README.md
@@ -96,9 +96,16 @@ You may need to delete the line: `themesDir = "../.."`
 
 To enable comments, add following to your config file:
 
-- DISQUS: `disqusShortname = YOURSHORTNAME`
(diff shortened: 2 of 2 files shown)


layouts/_default/single.html
```diff
@@ -51,7 +51,7 @@ <h1 class="f1 athelas mt3 mb1">
       {{- .Content -}}
       {{- partial "tags.html" . -}}
       <div class="mt6 instapaper_ignoref">
-      {{ if .Site.DisqusShortname }}
+      {{ if .Site.Config.Services.Disqus.Shortname }}
```
regisphilibert, line 54: Hi @jmooring and thank you for this! 

I wonder if there is a way to test both values as we'd like the theme to be retro compatible for some months at least.

Maybe just `{{ with .Site.DisqusShortname | or .Site.Config.Services.Disqus.Shortname }}` ?

regisphilibert (changes requested):

layouts/_default/single.html
jmooring: .Site.Config.Services.Disqus.Shortname looks for that config key, falling back to site.DisqusShortname.

jmooring: v0.41
<https://github.com/gohugoio/hugo/commit/4ddcf52ccc7af3e23109ebaac1f0486087a212ba>

`.Site.Config.Services.Disqus.Shortname` falls back to `disqusShortname`
`.Site.Config.Services.GoogleAnalytics.ID` falls back to `googleAnalytics`

and added in v0.55.0...

`.Site.Config.Services.RSS.Limit` falls back to `rssLimit`

Issue #502, opened 2022-01-31
Footer not at bottom when post is short 


rmx90210:
Hello,

Thank you for this lovely theme! I noticed one issue I wanted to surface here. When a post is very short (for example, just a sentence or two), the footer isn't at the bottom. Here is a screenshot of what I'm seeing:

![image](https://user-images.githubusercontent.com/1463553/151744637-dc5a76e0-4d1c-4c78-9403-29ecae4290a4.png)

I think this may be the same issue as #370, which appears to have been closed.

Altonss:
This trick resolved the issue:

```
body{
  min-height: 100vh;
  display: flex;
  flex-direction: column;
}

footer{
  margin-top: auto;
}
```

Altonss:
> Hi @Altonss, Thanks for your code snippet! Would you mind giving some details about where/in which file this code should be?

I added this in a `custom.css` file! I didn't know how if it is possible to modify the tachyon css syntax in the html templates to match this, so I was not able to propose this change in a PR :)
I would be happy to read a solution if anyone has an idea!

t-kimber:
> I added this in a `custom.css` file! I didn't know how if it is possible to modify the tachyon css syntax in the html templates to match this, so I was not able to propose this change in a PR :) I would be happy to read a solution if anyone has an idea!

Wonderful, thanks for your help @Altonss !

Juanpam:
@Altonss I don't want to steal that snippet and make a PR. You should get the credit for that one!

pythoninthegrass:
New-ish to hugo so this took a bit of troubleshooting. Got @Altonss' `custom.css` working with two files:
### `static/css/custom.css`
```css
body {
  min-height: 100vh;
  display: flex;
  flex-direction: column;
}

footer {
  margin-top: auto;
}
```
### `hugo.toml`
```toml
baseURL = 'https://example.org/'
languageCode = 'en-us'
title = 'Hello, World!'
theme = 'ananke'

[params]
custom_css = ["css/custom.css"]
custom_js = ["js/custom.js"]
```
Thanks for posting the workaround!! 🙏

PR #670, opened 2024-01-06
Limit footer to bottom

LageTs:
Hello everyone!

I also discovered what is already topic in https://github.com/theNewDynamic/gohugo-theme-ananke/issues/502 and wanted to fix this problem:

When the content of a page is so small that the full page is smaller than the view-port the footer is not at the bottom.

With those little changes the page will be at least the size of the view-port and the space between content and footer is automatically a fitting margin at the top of the footer.

Same idea was also mentioned by https://github.com/Altonss

### DIFF
--- a/assets/ananke/css/_styles.css
+++ b/assets/ananke/css/_styles.css
@@ -18,3 +18,13 @@
 .nested-links a{
   overflow-wrap: break-word;
 }
+
+body {
+  display: flex;
+  flex-direction: column;
+  min-height: 100vh;
+}
+
+footer {
+  margin-top: auto;
+}


ramiro:
Perhaps worth mentioning is that for this issue to manifest (or at least to increase the chance of this) there should be no `featured_image` defined that pushes the post body to the bottom.

e.g. commenting this out in the example site: https://github.com/theNewDynamic/gohugo-theme-ananke/blob/f34c219436ab1c3b116f8706a521d40c4b8693b9/exampleSite/content/en/_index.md?plain=1#L8

PR #675, opened 2024-01-29
Multilingual support: Add hreflang HTML tags

dogweather:
This implements Google's advice for webmasters: 
* https://developers.google.com/search/docs/specialty/international/localized-versions

I adapted the code from here:
* https://gohugo.io/methods/page/alltranslations/


dogweather:
@regisphilibert Does it look good to you?

Issue #711, opened 2024-09-09
bug: .Site.Social deprecated in Hugo v0.124.0


henry63:
With hugo 0.134, and anake 33fbda0 the file ananke/layouts/partials.social-share.html cause a warning :

    `WARN  deprecated: .Site.Social was deprecated in Hugo v0.124.0 and will be removed in a future release. Use .Site.Params instead.`
This happens even if this code is not present 
   ```
[[params.ananke_socials]]
name = "twitter"
```
This is probably caused by line 6,7,8  of  ananke/layouts/partials.social-share.html
```
{{ with site.Social.twitter }}
  {{ $twitter_href = printf "%s&via=%s" $twitter_href . }}
{{ end }}

```

**Comment by davidsneighbour** [comment]
Hi @henry63. Please have a look at #715 for a quick solution.

- add a `username = "your_twitter_username"` in your Twitter configuration under `params > ananke_socials`
- copy the [full file](https://raw.githubusercontent.com/theNewDynamic/gohugo-theme-ananke/41d183a03456470e3f8f01ff1298e1662fca1ee4/layouts/partials/social-share.html) into `layouts/partials/social-share.html` in your project folder (NOT in the theme folder) to override the current state of the theme.

I might not merge this fix (yet) because the integration of social media profiles should be overhauled, which is a breaking change. I'll update this issue report when there is an update.

**Comment by jmooring** [comment]
Prior to v0.136.0 this emits a warning.
With v0.136.0 this emits an error and fails the build.

**Comment by c0r0n3r** [comment]
Please consider #736

**Comment by davidsneighbour** [comment]
https://github.com/theNewDynamic/gohugo-theme-ananke/milestone/8

It says "due by Oct 20th," and I hope to meet that goal. We are behind with many new features, but deprecations come first, and fixing the overall social media network/icon implementation comes next. With the next release all "deprecated" warnings should be dealt with, and it shouldn't come to errors from then on (if GoHugo doesn't release so many new releases in a week that it overruns us).

**Comment by melroy89** [comment]
Now the build fails: 

```
ERROR deprecated: .Site.Social was deprecated in Hugo v0.124.0 and will be removed in Hugo 0.137.0. Implement taxonomy 'social' or use .Site.Params.Social instead.
Total in 926 ms
Error: error building site: logged 1 error(s)
```

yorffuoj:
[...]

PR #715, opened 2024-09-19
theme(fix): deprecated: .Site.Social

davidsneighbour:
Quick fix for the error "deprecated: .Site.Social" in #711 

This will not make it into master in this current format. Keep an eye on the issue to find out when we have a proper solution.


melroy89:
Plz merge.. This is now breaking the hugo build actually

PR #728, opened 2024-10-03
Fixed German translation

Pta42:
Fixed only some translation issues in the German translation. There were some "-" missing and I changed the "Dein Name" to "Ihr Name", which is more formal.

### DIFF
--- a/i18n/de.toml
+++ b/i18n/de.toml
@@ -20,16 +20,16 @@ other = "Was ist in dieser {{ .Type }}"
 other = "Ähnliches"
 
 [yourName]
-other = "Dein Name"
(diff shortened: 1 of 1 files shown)


davidsneighbour:
I think it would be best to have both versions (formal/informal) of these kind of languages included. `de.toml` and `de-formal.toml` should work. I will find out if there is a precedent on how to implement this without putting too much configuration work on the end-user. 

That might be something that needs to be brought upstream to find a clean solution. For now I will not merge this PR as we don't have any statistics as to who is using formal or informal language more in our userbase. 

On an individual base one can always override i18n files in the local repository.

PR #755, opened 2024-10-21
Fix social share cache

melroy89:
Do NOT use the same cache name for social _share_ and social _follow_..

This will never work of course.


davidsneighbour (approved):

melroy89:
You're welcome 😜🤗

melroy89:
I actually think it's still broken. Since the cache can't be used across multiple articles. You might need to remove the cache fully!

melroy89:
> I actually think it's still broken. Since the cache can't be used across multiple articles. You might need to remove the cache fully!

Found the problem, see PR: https://github.com/theNewDynamic/gohugo-theme-ananke/pull/757

PR #652, opened 2023-09-17
Change to /intent/tweet (Twitter share link)

melroy89:
Change from /share to /intent/tweet.

More info: https://developer.twitter.com/en/docs/twitter-for-websites/tweet-button/overview

PR #756, opened 2024-10-22
Move to x.com

melroy89:
Follow-up after: https://github.com/theNewDynamic/gohugo-theme-ananke/pull/652

### DIFF
--- a/config/_default/params.toml
+++ b/config/_default/params.toml
@@ -197,11 +197,11 @@ params = "posttype=link"
 
 [[ananke.social.networks]]
 slug = "twitter"
-label = "Twitter"
+label = "X"
 color = "#1da1f2"
-profile = "https://twitter.com/%s"
+profile = "https://x.com/%s"
 icon = "twitter" # font awesome brand icon name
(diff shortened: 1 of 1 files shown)


config/_default/params.toml
```diff
@@ -197,11 +197,11 @@ params = "posttype=link"
 
 [[ananke.social.networks]]
 slug = "twitter"
-label = "Twitter"
+label = "X"
```
davidsneighbour, line 200: Leaving that at Twitter is an opiniated decision. There is a x-twitter config further down the networks list that is labelled X. The changes to the share link are probably required (down under x-twitter too).

davidsneighbour (changes requested):

config/_default/params.toml
melroy89: Owh I see. That is very confusing to also have an x-Twitter option. I would just call it x.

davidsneighbour: I wanted it to be longer, and x-twitter is the slug Fontawesome is using... But I am flexible. Put it in the PR... one character slugs might lead to issues.

PR #757, opened 2024-10-22
Do not cache share links

melroy89:
You can't just cache the share links, that will result into every blog post having the same URL.

Just use `partial` on the getShareLink.html. So now each post will show the correct URL to share the post via social media or alike.

### DIFF
--- a/layouts/partials/social/share.html
+++ b/layouts/partials/social/share.html
@@ -34,7 +34,7 @@
         {{/* @todo notification into CLI that a network is configured but not supported */}}
       {{- end -}}
       {{- $options := (dict "context" page "setup" $setup) }}
-      {{- $href := partial "func/getShareLink.html" $options $options -}}
+      {{- $href := partial "func/getShareLink.html" $options -}}
       <a href="{{ $href }}"
         class="ananke-social-link {{ $setup.slug }} no-underline"
         title="{{ $label }}" aria-label="{{ $label }}"
--- a/layouts/partials/social/share.html
+++ b/layouts/partials/social/share.html
@@ -34,7 +34,7 @@
         {{/* @todo notification into CLI that a network is configured but not supported */}}
       {{- end -}}
       {{- $options := (dict "context" page "setup" $setup) }}
-      {{- $href := partialCached "func/getShareLink.html" $options $options -}}
+      {{- $href := partial "func/getShareLink.html" $options $options -}}
       <a href="{{ $href }}"
         class="ananke-social-link {{ $setup.slug }} no-underline"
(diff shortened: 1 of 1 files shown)


layouts/partials/social/share.html
```diff
@@ -34,7 +34,7 @@
         {{/* @todo notification into CLI that a network is configured but not supported */}}
       {{- end -}}
       {{- $options := (dict "context" page "setup" $setup) }}
-      {{- $href := partialCached "func/getShareLink.html" $options $options -}}
+      {{- $href := partial "func/getShareLink.html" $options $options -}}
```
melroy89, line 37: Why is there two times `$options`?

melroy89: Since we now use `partial` instead of `partialCached`, I believe we can just pass 1x the `$options`.

davidsneighbour:
False. There is a bug in the logic of the partial, that is true. But we CAN and will cache those links. A share link must be cacheable for a specific page. Check the $options dict. `page` here might refer to the outer list page a post is listed on, which is what ruins the cache if you implement share links on list pages. What this needs is `$context` (defined in the first line) instead of `page` in the options and then they (should be/are) individual caches per post object (if `$context` doesn't walk into the same trap). I'll need to check that though, it might be that I will inline the sharelink generation.

layouts/partials/social/share.html
```diff
@@ -34,7 +34,7 @@
         {{/* @todo notification into CLI that a network is configured but not supported */}}
       {{- end -}}
       {{- $options := (dict "context" page "setup" $setup) }}
-      {{- $href := partialCached "func/getShareLink.html" $options $options -}}
+      {{- $href := partial "func/getShareLink.html" $options -}}
```
davidsneighbour, line 37: nope ;) see my comment. `{{- $options := (dict "context" $context "setup" $setup -}}` should do the trick with unchanged `partialCached` line.

davidsneighbour (changes requested):

layouts/partials/social/share.html
melroy89: Sorry what? I don't see your comment.

davidsneighbour:
Something along the lines of this, not tested yet:

https://github.com/theNewDynamic/gohugo-theme-ananke/pull/763

davidsneighbour (approved):

davidsneighbour:
I am going to merge this now. I opened an issue for going to look at caching after the release.

https://github.com/theNewDynamic/gohugo-theme-ananke/issues/764

PR #758, opened 2024-10-22
Add transform.Plainify to the summary body content of the share links

melroy89:
You definitely want to **add Plainify transformer first**, because the body content will become a HTML encoding mess. I left `HTMLEscape` there as well, just be sure for security reasons. But if you want, I can remove the second transformer.

Like a tweet on X will look like this:

image

After implementing `transform.Plainify`, it will look much better:

image

(ignore the red, I don't have X premium pro blabla, it's red because I hit the max limit of chars)

### DIFF
--- a/layouts/partials/social/share.html
+++ b/layouts/partials/social/share.html
@@ -66,7 +66,7 @@
   {{- $setup := .setup -}}
   {{- $separator := "&" -}}
   {{- $title := $context.Title | transform.HTMLEscape -}}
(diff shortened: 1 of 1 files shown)


davidsneighbour:
The HTMLEscape is there so that special characters are escaped. There is a function [`urlize`](https://gohugo.io/functions/urls/urlize/) that might be suited better than that? I'll merge it and add some testing for it later.

PR #795, opened 2024-11-17
Fixes #794 - add show_recent_posts parameter to stop showing recent posts on index

### LINKED ISSUE #123: Allow multiple sections to be listed under recent
I'd like to have two sections listed under my home page just like the default "recent posts". Namely a "stories" and "posts" lists. If this would be implemented it would be great!

### LINKED ISSUE #794: bug: ananke will always show recent posts on index
This seems to be a regression, either because of the recent ananke changes or because of changes in hugo.
Setting `recent_posts_number = 0` will no longer disable showing recent posts on index.

I tested a little and I think this is because `{{ $n_posts := $.Param "recent_posts_number" | default 3 }}` will default to 3 if recent_posts_number is 0.

## Expected Behavior
Ability to disable recent posts on index

## Current Behavior
not possible to disable recent posts on index

## Possible Solution
Add a parameter to toggle showing recent posts

## Steps to Reproduce

1. set `recent_posts_number = 0`
2. observe recent posts are still being shown

## The bigger picture (we need context)

<details><summary>Hugo Config</summary>
<p>

</p>
</details>

Klaas-:
Closes #794

You can no longer set `recent_posts_number = 0` because in current hugo versions `{{ $n_posts := $.Param "recent_posts_number" | default 3 }}` will default to three if recent_posts_number is 0. This used to work a couple of versions back, I am assuming this is not a regression but rather wanted behavior in hugo -- so I introduce a new parameter (boolean) `show_recent_posts` to enable(default) or disable recent posts on index.

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -4,7 +4,7 @@
     </article>
     {{/* Define a section to pull recent posts from. For Hugo 0.20 this will default to the section with the most number of pages. */}}
     {{ $mainSections := .Site.Params.mainSections | default (slice "post") }}
-    {{ $show_recent_posts := .Site.Params.show_recent_posts | default (true) }}
+    {{ $show_recent_posts := site.Params.ananke.show_recent_posts }}
     
     {{/* Check to see if the section is defined for ranging through it */}}
     {{range ($mainSections)}}
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -4,6 +4,7 @@
     </article>
     {{/* Define a section to pull recent posts from. For Hugo 0.20 this will default to the section with the most number of pages. */}}
     {{ $mainSections := .Site.Params.mainSections | default (slice "post") }}
+    {{ $show_recent_posts := .Site.Params.show_recent_posts | default (true) }}
     
     {{/* Check to see if the section is defined for ranging through it */}}
     {{range ($mainSections)}}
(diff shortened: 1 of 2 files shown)


layouts/index.html
```diff
@@ -4,6 +4,7 @@
     </article>
     {{/* Define a section to pull recent posts from. For Hugo 0.20 this will default to the section with the most number of pages. */}}
     {{ $mainSections := .Site.Params.mainSections | default (slice "post") }}
+    {{ $show_recent_posts := .Site.Params.show_recent_posts | default (true) }}
```
davidsneighbour, line 7: Please prefer (for future compatibility) `site.Params` (global) to `.Site.Params` (based on the current context, that might change) and move the config under `ananke`. 

```
{{ $show_recent_posts := site.Params.ananke.show_recent_posts | default true }}
```

Also add the following to `config/_defaults/params.toml` in the beginning:

```
[ananke]
show_recent_posts = true # show recent posts on the homepage
```

Klaas-: if I set it as default in config/_defaults/params.toml I no longer need "default true" in the index.html, right? I'll update the PR

davidsneighbour (approved):

layouts/index.html
davidsneighbour: yes, the setting in the themes config will be default and then overridden by local settings.

PR #801, opened 2024-11-23
Support reading_speed language param in single.html

### LINKED ISSUE #123: Allow multiple sections to be listed under recent

D14rn:
This PR makes it possible to configure the reading speed in specific languages by adding a configuration like the one in the https://gohugo.io/methods/page/readingtime/ method page

example configuration that will be supported:
```toml
[languages.en.params]
reading_speed = 256
```

It doesn't necessarily solve an important problem because you can just override the theme's single.html file to achieve the same result, but maybe it would be nice to support reading_speed configuration out of the box.

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -43,7 +43,14 @@
           3) A page front matter value is set `show_reading_time = true`
         */}}
       {{ if (or (eq (.Param "show_reading_time") true) (eq $section.Params.show_reading_time true) )}}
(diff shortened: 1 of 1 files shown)


davidsneighbour:
I added this in 39e214598595 - there were too many changes to the layout to properly merge it without conflicts.

PR #816, opened 2025-01-15
i18n: Create Occitan translation

ensag-dev:
Hello,
I hope this PR is well formed

Regards

### DIFF
--- a/i18n/oc.toml
+++ b/i18n/oc.toml
@@ -1,0 +1,49 @@
+[more]
+other = "Mai"
+
+[allTitle]
+other = "Tot {{.Title }}"
+
+[recentTitle]
+other = "Recent {{.Title }}"
+
+[readMore]
+other = "ne legir mai"
+
+[by]
+other = "Per"
+
+[whatsInThis]
+other = "De qué i a dins {{ .Type }}"
+
+[related]
+other = "Ligats"
+
+[yourName]
+other = "Vòstre nom"
+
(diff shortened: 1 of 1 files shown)


davidsneighbour:
It looks ok. Two notes:

- I am not sure if `oc` is the right language code or if we will have to use some form of `ES_oc` - will investigate. That's just to go sure and due to my own knowledge deficits. This is connected to ISO 639-1, but it's more about working in connection with Hugo.
- You have a "Recent" there in the third group (line 8). Is that because there is no wording for that or did you miss it? I assume it's not Occitania?

Also, I am starting to attribute "code ownership" for translations to the people who translated it. Are you ok to be attributed with `ensag-dev` in .github/CODEOWNERS? The concept is explained [here in the docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners). Basically you own the translation, because no one else (except anyone adding changes later on) knows the language ;)

ensag-dev:
Hello,

the iso code would be oc_FR, oc_ES or oc_IT, the language code then the country.
Yes "recent" is the same in Occitan :)

davidsneighbour (approved):

davidsneighbour:
I am merging this. The RFC Hugo is using is [RFC5646](https://en.wikipedia.org/wiki/IETF_language_tag) and oc is in there.

PR #821, opened 2025-02-04
theme(fix): social follow/share using new config

### LINKED ISSUE #808: bug: social media network array override by custom network configuration
The table in `[[params.ananke.social.networks]]` is overridden by custom local configurations. 

See discussion in #805 

## Possible Solution

I think we need to rework the way custom social networks are integrated from the current solution:

```
[[params.ananke.social.networks]] 
slug = "wikipedia" 
label = "Wikipedia" 
color = "#0645ad" 
profile = "https://en.wikipedia.org/wiki/%s" 
icon = "wikipedia-w" 
```

to 

```
[params.ananke.social.networks.wikipedia]
label = "Wikipedia" 
color = "#0645ad" 
profile = "https://en.wikipedia.org/wiki/%s" 
icon = "wikipedia-w" 
```

where the slug is the key of the array. This makes it easy to override specific pre-defined networks and also define new items without breaking the pre-configured items.

jerr0328:
A recent commit changed the config for social to use a map instead of a list, this broke the social share/follow, though a quick fix seems to restore it and removes the need for sorting.

Closes #808

I checked the wikipedia example from the original issue and it seems like this will also fix that. I'm not sure if you had anything else to fix, feel free to push to my branch.


holywend:
I also got into similar situation, im new to hugo, and trying to make social works, all I can track back is the follow.html and share.html.

ill try to checkout this branch and see if it works!

thanks @jerr0328

davidsneighbour (approved):
looks good.

PR #848, opened 2025-05-27
Remove call to css.Sass

srwilson-test:
There is no associated issue. The css.Sass doesn't seem to exist anymore - my change seems to fix this. This could be an incorrect approach.


davidsneighbour:
1. Please do not open a PR without opening an issue or discussion first to make sure we have an actual issue. 
2. It is probably the other way around: [css.Sass](https://gohugo.io/functions/css/sass/) has existed since Hugo 0.128.0. 

Check your Hugo version with `hugo version` and upgrade to at least 0.128.0. If you can't upgrade then I suppose you always can downgrade to an older Ananke version, but you will run into other issues connected to old versions and deprecations. Please open a discussion if you need help with that.

PR #859, opened 2025-07-06
fix: update HTML language attributes and viewport meta tag for better responsiveness

poolboy17:


davidsneighbour:
This PR removes more than it adds. Everything between `<head>` and `</head>` is gone and three tags are re-added. This removed functionality from the `<title>`-tag. The only usable change is the content of the viewport meta tag that could enjoy an update.

PR #889, opened 2025-12-21
prior to ananke template replacement

monkeysh1ne:


davidsneighbour:
This is a personal preference, please create a local archetype file for your tags.

PR #908, opened 2026-01-22
feat: add commento.io frontend configuration

### LINKED ISSUE #875: feat: add commento.io frontend configuration
Add [commento.io frontend configuration](https://github.com/theNewDynamic/gohugo-theme-ananke/wiki/Customization-Comments#commentoio) for self hosted instances to the config object.

NITHINSPACETIME:
fixes #875

Added a `commentoFrontend` parameter to the config to make setting up self-hosted instances a lot easier for people ig.

If you set `commentoFrontend = "https://your-instance.com"`, it automatically points the script source there. If it's not set, it just falls back to the default CDN like before.

Also added a commented-out example in [params.toml](cci:7://file:///home/kamisama/Downloads/gohugo-theme-ananke/site/config/_default/params.toml:0:0-0:0) so people know it's there.


davidsneighbour:
I am actually in the process of removing this. I want instead add a `layouts/_partials/comments.html` that is empty by default and then in the docs have explanations for any comment system. 

Regarding the code: I find my solution (full path to commento.js) easier for more common users because they don't need to think about what URL to add and it won't come to issues with trailing slashes or paths. The `commento.js` file _might_ have a different filename. If a person wants to configure they configure that one full path to the JavaScript file instead of a domain and a path.

Lastly: your comment in the configuration is in `site.Params.ananke.home.commentoFrontend` and NOT `site.Params.commentoFrontend`.

I'll keep this open and we'll move it to the docs once that is implemented.

davidsneighbour:
This got somehow closed. Marking it `status:recheck` so it does not get lost. If you want to reopen please adapt to the new repo structure (github.com/gohugo-ananke/ananke) and pull against the development branch. If that is too much to fix please don't delete this PR and I'll adapt it.

PR #918, opened 2026-02-07
update deprecated pagination config

olafk:
to prevent the error message:

ERROR deprecated: site config key paginate was deprecated in Hugo v0.128.0 and subsequently removed. Use pagination.pagerSize instead.

As this is such a trivial change in a sample file, I didn't file an issue first, but just edited/PR'ed this file. If this is not ok and you need an issue: Please let me know.


olafk:
Not sure what to do about the failed deployment. I feel it's unrelated to the proposed change.

PR #933, opened 2026-04-14
fix: address deprecations in language API

jmooring:
See https://discourse.gohugo.io/t/deprecations-in-v0-158-0/.


davidsneighbour:
I am merging this. For future PRs please keep theme.toml and config/_default/hugo.toml updated regarding the min version. This is a bit chaotic currently (hugo.toml says 0.160, theme says 0.128). But for future releases I think it's best if every release is coupled to the latest version as min version to keep us to date with the latest versions and users on older Hugo releases need to use the version that was up to date at that point.

davidsneighbour (approved):

